PR #11, opened 2022-06-26
Physics

mat-1:
The player has physics and collisions.


mat-1:
physics isn't fully done but it's pretty close and there's stuff fixed in this branch that's not in master so i'm merging rn

PR #12, opened 2022-07-23
Palette resizing

mat-1:
Makes it so blocks can be changed in the world which is kind of important.


mat-1:
I didn't test the global palette but it probably works

PR #14, opened 2022-08-01
Better errors

mat-1:
Currently most errors are just `String`s, this makes them proper enum errors.


mat-1:
TODO: there's still some `Result`s that use String as the error type.

PR #17, opened 2022-08-08
NBT decoder optimizations

mat-1:
Use AHash for HashMaps and make read_string read everything at once. I'm sure more optimizations are possible here but I don't know how to execute them. I also know I could make it way faster if it was a &[u8] instead of a Read, but it is, so I have to figure out how to optimize with that.


mat-1:
i can't come up with better optimizations so I'm merging it like this

PR #29, opened 2022-10-14
Microsoft Authentication

mat-1:
Allows users to join online-mode servers.

Caching is currently not implemented


Sculas:
It seems that the encryption request handling is done in the wrong order:
https://github.com/mat-1/azalea/blob/cb001fa341636d1fa6cc186bfd372f092088270c/azalea-client/src/client.rs#L159-L187

`conn.authenticate(...)` expects `writer.private_key` to be set:
https://github.com/mat-1/azalea/blob/cb001fa341636d1fa6cc186bfd372f092088270c/azalea-protocol/src/connect.rs#L194-L203

but `writer.private_key` is set when `conn.set_encryption_key(...)` is called:
https://github.com/mat-1/azalea/blob/cb001fa341636d1fa6cc186bfd372f092088270c/azalea-protocol/src/connect.rs#L157-L164

but `conn.set_encryption_key(...)` is called last:
https://github.com/mat-1/azalea/blob/cb001fa341636d1fa6cc186bfd372f092088270c/azalea-client/src/client.rs#L186

and `conn.authenticate(...)` is called before:
https://github.com/mat-1/azalea/blob/cb001fa341636d1fa6cc186bfd372f092088270c/azalea-client/src/client.rs#L164-L171

mat-1:
> It seems that the encryption request handling is done in the wrong order:

Fixed :)

PR #31, opened 2022-10-17
refactor: remove println statements

Sculas:
This PR removes all println statements and logs them on the trace level instead.
Usually, libraries shouldn't print to stdout using println, since there's no control over them.

Some println statements also looked like leftover debug code, which I have removed.
I can add those back if requested.


mat-1 (approved):
Thank you, they are indeed debug leftover code that I forgot to remove :)

PR #33, opened 2022-10-21
Options

mat-1:
Adds Client::set_client_information and sends a packet to the server with it. If set_client_information isn't called before login then a default will be sent to the server, though the client can change the options after login.


mat-1:
Todo: add an option to the `azalea` crate to change the options

mat-1:
I decided that the best way to do it is to have an `Initialize` event that's fired immediately after the bot joins, so a new field doesn't have to be added to azalea::start

PR #34, opened 2022-10-30
1.19.3

mat-1:


mat-1:
22w43a works

mat-1:
(haven't tested 22w44a in-game but it didn't change anything in the protocol so it probably still works)

mat-1:
So it seems like the generator mod that's used for shape generation was broken in the new Fabric/Minecraft versions, too lazy to fix myself so I'll probably be switching to [Pixlyzer](https://gitlab.bixilon.de/bixilon/pixlyzer) for shape generation soon.

mat-1:
Haven't tested ingame

mat-1:
Codegen kept breaking but now I finally got it to generate everything. Haven't tested in-game though.

mat-1:
Tested in-game, works :)

PR #40, opened 2022-11-14
Military-grade server implementation

HonbraDev:
this would simplify `Connection<R, W>` grinders

### DIFF
--- a/azalea-protocol/src/connect.rs
+++ b/azalea-protocol/src/connect.rs
@@ -302,7 +302,7 @@ where
     R1: ProtocolPacket + Debug,
     W1: ProtocolPacket + Debug,
 {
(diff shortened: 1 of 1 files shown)


azalea-protocol/src/connect.rs
```diff
@@ -323,4 +323,24 @@ where
             },
         }
     }
+
+    pub fn wrap(stream: TcpStream) -> Connection<R1, W1> {
```
mat-1, line 327: LGTM, but can you add a doc comment for the function explaining what it does / what it's meant to be used for?

mat-1 (changes requested):

mat-1 (approved):

PR #41, opened 2022-11-18
Player List

mat-1:
Keep track of the players in the tab list in the Client::players field and send Event::UpdatePlayers(UpdatePlayersEvent) when the player list changes.


mat-1:
It should work, but I haven't tested in-game

PR #42, opened 2022-11-18
Add functions to get ChatPacket author and content

mat-1:


mat-1:
Haven't tested ingame

PR #46, opened 2022-11-28
Uploaded doc workflow

AtomicGamer9523:
Added a workflow to automatically document everything

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -28,7 +28,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
-      - run: cargo doc --all
+      - run: cargo doc --workspace --no-deps
       - uses: "finnp/create-file-action@master"
         env:
           FILE_NAME: "./target/doc/index.html"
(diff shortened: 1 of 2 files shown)


.github/workflows/doc.yml
```diff
@@ -0,0 +1,45 @@
...
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+      - run: cargo doc --all
```
mat-1, line 31:
```suggestion
      - run: cargo doc --workspace --no-deps
```
This seems to document every dependency and according to the cargo book --all is a deprecated alias for --workspace

mat-1 (changes requested):
Thank you!! I was actually trying to find out how to do this earlier but I couldn't find a good solution, so I very much appreciate this :)

mat-1 (approved):
LGTM
idk how to test it so i hope it works and if not I'll fix it

PR #47, opened 2022-12-03
Serialize Component

EightFactorial:
Also fixes ClientboundStatusResponsePacket


mat-1 (approved):
Thank you! I made my own commit to make it compile (since the "derive" feature wasn't enabled for Serde) and did some minor changes to make the code slightly more consistent with other things.

PR #50, opened 2022-12-06
Reauth on invalid session

mat-1:


mat-1:
I didn't test it super thoroughly but eh it should work.

PR #56, opened 2023-01-15
Add function to get message sender's UUID

EightFactorial:


mat-1 (approved):
Thanks, this doesn't work for servers that use system chat packets but ehh it's fine

PR #58, opened 2023-01-20
Allow using `azalea-chat` without `azalea-buf` to avoid unstable features

Nemo157:
We'd like to use `azalea-chat` in a custom bot to render colored text, but using the stable compiler (https://github.com/booleancoercion/ferrisquery/pull/1). We're not relying on the `azalea-buf` functionality for this so just disabling that and relying on the json serialization works fine for us. I'm not sure whether this is something you'd want to support upstream, we're fine just relying on a fork if not.

This is technically a breaking change for any users that had `azalea-chat.default-features = false` and were relying on the `azalea-buf`` functionality.

### DIFF
--- a/azalea-chat/Cargo.toml
+++ b/azalea-chat/Cargo.toml
@@ -8,10 +8,13 @@ version = "0.5.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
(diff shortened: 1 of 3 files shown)


mat-1 (approved):

mat-1:
Thank you! Breaking changes aren't an problem since all the crates are still in 0.x :)

PR #60, opened 2023-01-24
Fix test and packets

EightFactorial:


azalea-protocol/src/packets/game/clientbound_initialize_border_packet.rs
```diff
@@ -16,3 +16,19 @@ pub struct ClientboundInitializeBorderPacket {
...
 }
+
+impl McBufWritable for ClientboundInitializeBorderPacket {
+    fn write_into(&self, buf: &mut impl std::io::Write) -> Result<(), std::io::Error> {
+        self.new_center_x.write_into(buf)?;
+        self.new_center_z.write_into(buf)?;
+        self.old_size.write_into(buf)?;
+        self.new_size.write_into(buf)?;
+        self.lerp_time.var_write_into(buf)?;
+        // Odd
```
mat-1, line 27: This is very suspicious. Maybe the VarLong implementation is wrong?

azalea-protocol/src/packets/game/clientbound_player_info_update_packet.rs
```diff
@@ -196,6 +196,6 @@ impl McBufWritable for ActionEnumSet {
         if self.update_display_name {
             set.set(5);
         }
-        set.write_into(buf)
+        set.as_u8().write_into(buf)
```
mat-1, line 199: It's supposed to already have this behavior if there's at most 8 bits. This seems more like a bug with writing bitsets.

mat-1 (changes requested):
There's two fixes here that don't really seem like the "proper" fix. I'll investigate a bit to see if I can find the root causes.

azalea-protocol/src/packets/game/clientbound_initialize_border_packet.rs
EightFactorial: Oh, it is a bug with VarLong. It's supposed to write a zero into the buffer at the end.
Here's the function from PacketByteBuf:

public PacketByteBuf writeVarLong(long value) {
    while((value & -128L) != 0L) {
        this.writeByte((int)(value & 127L) | 128);
            value >>>= 7;
    }

    this.writeByte((int)value);
    return this;
}

mat-1 (approved):

PR #61, opened 2023-01-30
More packet fixes, tests, handle error

EightFactorial:


azalea-protocol/src/packets/game/clientbound_boss_event_packet.rs
```diff
@@ -116,28 +117,28 @@ pub struct Properties {
 
 impl McBufReadable for Properties {
     fn read_from(buf: &mut Cursor<&[u8]>) -> Result<Self, BufReadError> {
-        let byte = u8::read_from(buf)?;
+        let set = FixedBitSet::<3>::read_from(buf)?;
```
mat-1, line 120: This is definitely better than how it was before, though at some point I want to make it so there's just a derive macro that generates these.

mat-1 (approved):

PR #64, opened 2023-02-08
derive `Debug` for `BlockState`

ChemicalXandco:
When I want to know what block is at a position, I cannot because the `Debug` impl does not not produce the name of the block. With this change, I get the desired output.

### DIFF
--- a/azalea-block/Cargo.toml
+++ b/azalea-block/Cargo.toml
@@ -6,6 +6,9 @@ name = "azalea-block"
 repository = "https://github.com/mat-1/azalea/tree/main/azalea-block"
 version = "0.5.0"
 
+[features]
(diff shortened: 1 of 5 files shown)


mat-1:
This was intentionally not implemented since it slows down compile times too much and significantly increased the size of the binary. You can accomplish something similar with `Box::<dyn Block>::from(block_state).id()`. Maybe this could still be added but behind a feature flag though?

mat-1 (changes requested):
This was intentionally not implemented since it slows down compile times too much and significantly increased the size of the binary. You can accomplish something similar with `Box::<dyn Block>::from(block_state).id()`. Maybe this could still be added but behind a feature flag though?

mat-1 (approved):

PR #65, opened 2023-02-08
Add KeepAliveEvent and Fix Error Messages

EightFactorial:


azalea-world/src/entity/mod.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(clippy::derived_hash_with_manual_eq)]
+#![allow(clippy::derive_hash_xor_eq)]
```
mat-1, line 2: This was renamed in a newer nightly version.

azalea-client/src/packet_handling.rs
```diff
@@ -925,10 +943,21 @@ impl PacketReceiver {
...
+                Ok(packet) => {
+                    self.packets.lock().push(packet);
+                    // tell the client to run all the systems
+                    self.run_schedule_sender.send(()).await.unwrap();
+                }
+                Err(error) => {
+                    if let ReadPacketError::ConnectionClosed = *error {
+                        return;
+                    } else {
+                        panic!("Error reading packet from Client: {error:?}");
```
mat-1, line 957: Panics are bad since they'll completely crash any auto-reconnecting bot

mat-1 (approved):

PR #67, opened 2023-02-12
Better chat events

mat-1:
This basically makes it so you can send chat packets more easily from a plugin, and also moves around some stuff to make more sense.


mat-1:
pretty sure this works fine

PR #68, opened 2023-02-13
fix `BlockCollisions` bounding box

ChemicalXandco:
When the bot is moving it will sometimes think it is not on the ground which will cause it to move only very slowly. The coordinates of the bounding box should be floored because that gives the correct block position.


mat-1:
Hold on, does `as i32` not have the same behavior as `.floor()`? or is it only when it's negative or something? There should probably be tests for this too.

mat-1 (approved):
Thanks!

PR #82, opened 2023-03-11
Fix infinite recursion in the DNS resolver

LunaWasFlaggedAgain:
Previously, when resolving server addresses into IPs, the resolver would cause an infinite recursion if the SRV record pointed to its own domain: `_minecraft._tcp.example.com. 300 IN	SRV	30 0 25566 example.com.`.

This PR fixes that.


mat-1 (approved):
Thanks! This can probably still be exploited if a server chains redirects so in the future it should be changed to have a better loop prevention (like remembering the previously looked up addresses and also preventing too many redirects) but this is fine for now.

PR #84, opened 2023-03-25
implement macro

Altidias:


AtomicGamer9523:
LGTM

mat-1 (changes requested):
Thank you!

mat-1:
Just gonna close this for now. Tbh after thinking about it more I don't think the macro is really necessary since the code wasn't that complex in the first place, but the other cleanups look good.

PR #88, opened 2023-06-02
Add functions `auth_with_link_code`, `get_ms_link_code`, and `get_ms_auth_token`.

Mythbusters123:
This allows for the seamless ease of getting the Microsoft Link code (should change function name later) and allows the user to authenticate these accounts through the specific device code response. 

This is a preliminary implementation. Could probably improved and refactored a fair bit as there is a lot of reused code. Thank you for considering this pull request.

Work still needed to be done:

- [ ] Refactor Code (reduce code repetition, rename `user_code` to `link_code` for clarity)
- [x] Add coherent documentation with a way on how to construct an `Account` struct.


mat-1 (approved):
this is probably fine

PR #93, opened 2023-06-22
Add ClientBuilder:new_without_log

xtexChooser:
https://github.com/mat-1/azalea/issues/92

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ members = [
   "azalea-registry",
   "azalea-inventory",
 ]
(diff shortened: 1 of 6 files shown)


mat-1 (changes requested):
Tbh I think this would be better and simpler if it was behind a feature flag [like how Bevy does it](https://github.com/bevyengine/bevy/blob/main/crates/bevy_internal/src/default_plugins.rs#L56-L141)

mat-1 (approved):

PR #94, opened 2023-07-09
Update to Bevy 0.11

mat-1:


mat-1:
This probably won't break anything too badly, I'm considering doing a change that will break a lot of stuff though...

PR #99, opened 2023-08-04
1.20.2

mat-1:


mat-1:
23w31a works, didn't test too much but at least clients can stay in the world for more than 15 seconds without erroring

mat-1:
(haven't tested the latest merge+update yet but i think it should be fine)

mat-1:
well the tests fail so

mat-1:
1.20.2-pre4 works

mat-1:
lots of changes so i probably messed something up, i'll fix later if i did

PR #101, opened 2023-08-19
Export bot.rs

TheDudeFromCI:
Currently, nearly all of the code used in Bot.rs is hidden, despite being written as an API. This means that components like Bot and events like LookAtEvent are inaccessible.

I think these are probably supposed to be public.


mat-1 (approved):
yep, this was my bad, thanks

PR #102, opened 2023-08-19
Created nearest_entity system param

TheDudeFromCI:
This utility function (provided as a `SystemParam` for easier Bevy system integration) can be used to locate nearby entities using much less boilerplate code.

_Note that the example in this PR relies on #101 for sending LookAt events._

### DIFF
--- a/azalea/examples/nearest_entity.rs
+++ b/azalea/examples/nearest_entity.rs
@@ -1,0 +1,54 @@
+use azalea::nearest_entity::EntityFinder;
+use azalea::ClientBuilder;
+use azalea::{Bot, LookAtEvent};
+use azalea_client::Account;
+use azalea_entity::{EyeHeight, Position};
+use bevy_app::{FixedUpdate, Plugin};
+use bevy_ecs::{
+    prelude::{Entity, EventWriter},
+    query::With,
+    system::Query,
+};
+
+#[tokio::main]
+async fn main() {
+    let account = Account::offline("bot");
+
+    ClientBuilder::new()
+        .add_plugins(LookAtStuffPlugin)
+        .start(account, "localhost")
+        .await
+        .unwrap();
+}
+
+pub struct LookAtStuffPlugin;
+impl Plugin for LookAtStuffPlugin {
+    fn build(&self, app: &mut bevy_app::App) {
+        app.add_systems(FixedUpdate, look_at_everything);
+    }
+}
+
+fn look_at_everything(
+    bots: Query<Entity, With<Bot>>,
+    entities: EntityFinder,
+    entity_positions: Query<(&Position, Option<&EyeHeight>)>,
+    mut look_at_event: EventWriter<LookAtEvent>,
+) {
+    for bot_id in bots.iter() {
+        let Some(entity) = entities.nearest_to_entity(bot_id, 16.0) else {
+            continue;
+        };
+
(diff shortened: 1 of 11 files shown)


azalea/examples/nearest_entity.rs
```diff
@@ -0,0 +1,73 @@
...
+    system::Query,
+};
+
+#[tokio::main]
+async fn main() {
+    let account = Account::offline("bot");
+
+    ClientBuilder::new()
+        .add_plugins(LookAtStuffPlugin)
+        .start(account, "localhost:34071")
```
mat-1, line 20: I'm guessing this is because you opened to LAN and didn't change the port, but in 1.20 you can just specify the port to be 25565 when opening to LAN.

azalea/examples/nearest_entity.rs
```diff
@@ -0,0 +1,73 @@
...
+
+pub struct LookAtStuffPlugin;
+impl Plugin for LookAtStuffPlugin {
+    fn build(&self, app: &mut bevy_app::App) {
+        app.add_systems(FixedUpdate, (look_at_everything, log_nearby_item_drops));
+    }
+}
+
+fn look_at_everything(
+    bots: Query<Entity, With<Bot>>,
```
mat-1, line 33: Usually I go with `With<Local>`  instead of `With<Bot>` since `Bot` isn't included for bots made with azalea-client. Doesn't matter that much but I figure for an example it should be consistent with the others.

azalea/src/nearest_entity.rs
```diff
@@ -0,0 +1,181 @@
...
+/// This system parameter allows for additional filtering of entities based off
+/// of ECS marker components, such as `With<>`, `Without<>`, or `Added<>`, etc.
+/// All functions used by this system parameter instance will respect the
+/// applied filter.
+///
+/// ```no_run
+/// /// All bots near aggressive mobs will scream in chat.
+/// pub fn bots_near_aggressive_mobs(
+///     bots: Query<Entity, With<Bot>>,
+///     entity_finder: EntityFinder<With<Aggressive>>,
```
mat-1, line 21: `Aggressive` is a metadata thing that's only used for like aggressive wolves and stuff iirc, this should probably just use `AbstractMonster` instead.

azalea/src/nearest_entity.rs
```diff
@@ -0,0 +1,181 @@
...
+
+/// This system parameter can be used as a shorthand for quickly finding an
+/// entity, (or several) close to a given position.
+///
+/// This system parameter allows for additional filtering of entities based off
+/// of ECS marker components, such as `With<>`, `Without<>`, or `Added<>`, etc.
+/// All functions used by this system parameter instance will respect the
+/// applied filter.
+///
+/// ```no_run
```
mat-1, line 17: no_run isn't necessary here since it's defining a function and the function isn't called.

mat-1 (changes requested):
This is great, thanks. Btw rn the examples and doctests fail when you run `cargo test`.

azalea/src/nearest_entity.rs
TheDudeFromCI: Turns out I need the `no_run` tag because of the imports. Otherwise the compiler complains about the missing imports used in the example.

mat-1 (dismissed):

mat-1 (approved):

PR #104, opened 2023-08-19
Removed .vscode settings

TheDudeFromCI:
Usually, this folder is added to the git ignore since these settings are often customized per-user. I have edited the `settings.json` on my own local fork of the project as well based on my own extensions and utilities.


mat-1 (approved):
thanks!

PR #105, opened 2023-08-23
Remove "despawning entity" spam

MDevVal:
This likely wasn't intentional


mat-1:
thanks, i replaced it with a `trace!` in https://github.com/mat-1/azalea/commit/57e5a0f0b96a38674bd18ac38d2d07e4f4ca2fd6

PR #107, opened 2023-08-30
Codegen rewrite

katietheqt:
This is currently a draft PR - all commits should be buildable but are not nicely divided by feature. I will squash before merging.


mat-1:
I'm gonna close this for now since I don't want to leave this PR open forever. My main complaints (mentioned in Discord dms) were that I don't want the codegen to be doing the disassembly directly since it's annoying to maintain, and that using Pixlyzer (or something similar) is kind of a requirement since block shapes are generated in ways that aren't easy to extract without a game mod.

Also to explain some things about the directory structure, the `gen*.py` files are for manually rerunning parts of the codegen for when the codegen code is modified, and the short names like "lib" and "code" are so the imports don't get annoying to type. Most of the refactoring changes seem fine though.

PR #110, opened 2023-10-08
1.20.3

mat-1:


mat-1:
(pretty sure at least FormattedText serialization/deserialization is broken, haven't tested in-game yet)

mat-1:
Block states are still broken because Mojang did something weird with pumpkins and melons. This is an issue in Burger.

PR #112, opened 2023-11-29
fixes trust_dns_resolver issue

1zun4:
Hi there,

For example, when trying to resolve `play.pika-network.net` it failed with the error: "Label contains invalid characters".
By using `Name::from_asciii(...)` this problem is solved and we can resolve the domain successfully.


mat-1 (approved):
LGTM, thanks!

Issue #98, opened 2023-07-28
Socks5 proxy support for client


NiiightmareXD:

1zun4:
Can be closed now.

PR #113, opened 2023-11-29
socks5 support

1zun4:
Introduces socks5 support to azalea. I am using this myself and it works well.
Fixes https://github.com/azalea-rs/azalea/issues/98

Might not be ideally implemented.


azalea-protocol/src/connect.rs
```diff
@@ -257,15 +258,43 @@ pub enum ConnectionError {
     Io(#[from] std::io::Error),
 }
 
+#[derive(Debug, Clone)]
+pub struct Proxy {
+    pub addr: SocketAddr,
+    pub auth: Option<(String, String)>,
```
mat-1, line 264: This should be a socks5_impl::protocol::UserKey instead of (String, String)

mat-1 (changes requested):
Thank you for writing this!

I think most of the functions that take in Proxy as an argument (`Client::join`, `ping_server`, `Connection::new`, `ClientBuilder::start`, `Swarm::add`, `Swarm::add_with_exponential_backoff`) should be split into separate functions to keep them simple since it's a lot more common for the user to not be using a proxy (and because I might want to add a feature gate to optionally disable the socks5-impl dependency in the future).
For example `Connection::new` could be split into `new(address: &SocketAddr)`, `new_with_proxy(address: &SocketAddr, proxy: Option<Proxy>)`, and `new_from_stream(stream: TcpStream)`.

1zun4:
> Thank you for writing this!
> 
> I think most of the functions that take in Proxy as an argument (`Client::join`, `ping_server`, `Connection::new`, `ClientBuilder::start`, `Swarm::add`, `Swarm::add_with_exponential_backoff`) should be split into separate functions to keep them simple since it's a lot more common for the user to not be using a proxy (and because I might want to add a feature gate to optionally disable the socks5-impl dependency in the future). For example `Connection::new` could be split into `new(address: &SocketAddr)`, `new_with_proxy(address: &SocketAddr, proxy: Option<Proxy>)`, and `new_from_stream(stream: TcpStream)`.

I completely agree with your suggestion to split the functions that take a Proxy as an argument into separate functions for simplicity, such as adding a feature gate to optionally disable the socks5-impl dependency.

I initially implemented it in a quick-and-dirty manner for the sake of expediency, but your feedback is spot-on. Expect these changes soon.

mat-1:
I don't mean to bother you, but do you still plan on finishing up this PR? If not then I'd be willing to finish it myself.

1zun4:
> I don't mean to bother you, but do you still plan on finishing up this PR? If not then I'd be willing to finish it myself.

Go ahead. I was going to, but I don't have enough time at the moment.

mat-1:
Merged in https://github.com/azalea-rs/azalea/commit/353eda21ac8280213edcec3823cc3bd77fe17c44 :)

I did a few changes: separating the socks5 and non-socks5 functions, making it easier to set which proxy a bot will use and removing the randomization, and generally cleaning it up a little so things like setting a custom join address for a bot and its proxy is set from a single options struct

PR #114, opened 2023-11-29
added disconnect reason

1zun4:
Hello there,

with this PR I have added the possibility to catch disconnect events and their reason easily.

But here is an issue that I am facing and cannot figure out how to fix:
When the server disconnects us from the game, we normally get `ClientboundGamePacket::Disconnect` but the connection is closed before we can read the packet and fire the event with an reason. This causes us to not get the actual disconnect message and I really need HELP with this issue.

### DIFF
--- a/azalea-client/src/client.rs
+++ b/azalea-client/src/client.rs
@@ -422,6 +422,7 @@ impl Client {
     pub fn disconnect(&self) {
         self.ecs.lock().send_event(DisconnectEvent {
             entity: self.entity,
+            reason: None,
         });
     }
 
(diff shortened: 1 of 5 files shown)


1zun4:
How it should look like (disconnect by Velocity).
image

How it fails (disconnect by plugin):
image

Seems Velocity, when disconnected by plugin, sends it as `ClientboundCustomPayloadPacket`, that explains why I could not capture it as `ClientboundGamePacket::Disconnect`.

mat-1 (approved):
Thanks! I'm changing the `reason` to an `Option<FormattedText>` after merging this though since there's some cases where the user will want to format the text another way. Also I noticed a couple bugs with disconnecting while testing this but those are totally my fault.

PR #115, opened 2023-11-29
Serde support for 'ServerAddress'

1zun4:
Hey there,

this PR adds support for serde serialization and deserialization for the ServerAdress struct. I find this quite useful for my configuration purposes and because this library already features serde, might add this support.


mat-1 (approved):
Thanks!

PR #118, opened 2023-12-04
Add masked chat type/event

ShayBox:


mat-1 (approved):
Thanks! I'm renaming the packet back to Disguised and simplifying the split_sender_and_content implementation after merging this.

PR #122, opened 2023-12-10
Add mining to the pathfinder

mat-1:


mat-1:
this is pretty much done i think, i want to add some more utilities to the pathfinder later (like infinite mining and mining an area) but that should be in a separate pr (and probably also a plugin separate from PathfinderPlugin, ig it could just be called PathfinderUtilsPlugin)

PR #123, opened 2023-12-16
Pathfinder extras

mat-1:
Adds a new internal plugin with more fun pathfinder things, like mining out an area.

### DIFF
--- a/azalea-client/src/chunks.rs
+++ b/azalea-client/src/chunks.rs
@@ -71,7 +71,7 @@ pub struct ChunkBatchFinishedEvent {
     pub batch_size: u32,
 }
 
-fn handle_receive_chunk_events(
+pub fn handle_receive_chunk_events(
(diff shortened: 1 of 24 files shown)


kittenvr:
is this like this pathfinding? https://cdn.discordapp.com/attachments/990193314496131113/1232274074504396820/Screen_Recording_2024-04-22_at_7.04.32_PM_2.mp4?ex=663216a9&is=6630c529&hm=3b4e2dabccd57a20abdc34ceca071a7f0d21ea4ff0b061d3bb8f019d99f4e783&

GaspardCulis:
Is it possible to add an event that gets sent when pathfinding is over? I haven't found clean ways to execute a system after pathfinder reached is goal

mat-1:
This ain't getting merged (but parts of it have slowly been getting implemented in the main branch over time.)

PR #124, opened 2023-12-17
Publicize all fields on `AddOperation` and `Style`

112batman:
The struct itself was already public, I assume the fields where meant to be as well.


mat-1 (approved):
Yep, they should be public. Thanks.

PR #125, opened 2023-12-17
Add missing particles

112batman:
Has all particles present on https://wiki.vg/Protocol#Particle as of 17 December 2023.
This also fixes particles being parsed as the incorrect type due to id shifts caused by newly added particles.

Please squash.


mat-1 (approved):
Thanks! The ParticleData enum should probably be changed to fail to compile when it doesn't match up with azalea_registry::ParticleKind (since that one is updated automatically) but that can be done later.

PR #127, opened 2024-01-18
1.20.5

mat-1:


mat-1:
Haven't tested 1.20.5-rc1 yet but the changes are pretty trivial so it should still work.

mat-1:
i realized the tags are getting shuffled on every update which isn't ideal but doesn't actually matter lol. also i haven't tested this version specifically but i tested the last release candidate where they made protocol changes so it should be good to merge.

Issue #134, opened 2024-03-08
Auth without remote connect? (feature request/inquiry)


JewFrags:
I was wondering if it's currently possible to authenticate/login to an account using the access token/email & passwd without needing to enable remote connect (linking with the code) otherwise I'd like to suggest this as a possible addition.

mat-1:
Yep, this is possible by manually creating the [Account](https://azalea.matdoes.dev/azalea/struct.Account.html).

PR #137, opened 2024-03-17
Generic Auth Flows

veronoicc:
Trying to make generic auth flows work, meaning that you can also use custom authentication servers.
Ik the code is horrible and missing documentation, feel free to help me out


veronoicc:
If i'm thinking correctly this would also fix #134

AS1100K:
I am not sure, I may be wrong. IDK what you mean by Generic Auth flow, but i would assume that you can use `username` and `password` to login without dealing with tokens. I think adding this is quite easy but seeing the number of changes you did i.e. 25 files seems a lot to me.
Here is an [example](https://learn.microsoft.com/en-us/entra/identity-platform/scenario-desktop-acquire-token-username-password?tabs=dotnet) for getting `access_token` via `username` and `password`. And just creating a function in `Account` lets say `microsoft_via_username` and this function will get the `access_token` and then follow `microsoft` method for further auth.

mat-1:
Closing since this PR is really old. Sorry that I never wrote a proper review.

I like the goal here; I've seen several people bring up this issue that creating `Account`s from tokens or with custom auth servers is too annoying.

If you (or anyone else) wants to pick this up, the main things I'd change in this PR are:
- Avoiding generic arguments (functions like `add_account` should take a boxed account instead so the user can mix accounts more easily).
- Rename `BoxedAccount` to just `Account` and the account trait probably to `AccountTrait`, since the boxed account will be the one that the user will probably be interacting with the most. The boxed account should probably also get a Deref impl.
- Preserving the existing `Account::offline` and `Account::microsoft` functions to avoid breaking changes and since imo these look nicer than `OfflineAccount::new` / `MicrosoftAccount::new`.
- I think keeping the old `fetch_certificates` return type of `Result<Certificates, FetchCertificatesError>` and adding like an `Unsupported` variant to the error is more clear than returning `Result<Option<Certificates>, FetchCertificatesError>`.

PR #139, opened 2024-03-22
fix: Add missing `successful` field to `ServerboundCustomQueryAnswerPacket`

Sculas:
While working on my own Minecraft server (using azalea) and reading [wiki.vg/Protocol](https://wiki.vg/Protocol) a lot, I noticed that the `successful` field is not implemented in the `ServerboundCustomQueryAnswerPacket`, as described here: [wiki.vg/Protocol#Login_Plugin_Response](https://wiki.vg/Protocol#Login_Plugin_Response).

This PR adds the missing field and removes an unused packet (that's not mapped or used anywhere).


Sculas:
Whoops - just realized I was looking at the wrong packet in `wiki.vg`!

PR #146, opened 2024-05-09
Bevy App brigadier example

EightFactorial:
Adds an example for how to use brigadier inside a bevy App.

Also fixes an issue where the `azalea_buf` feature was not enabled for `azalea_chat`.


mat-1 (approved):
lgtm, thanks

PR #147, opened 2024-05-14
add as_property_list method for Block

urisinger:
this method is supposed to make it easier to query block properties in a generic non block specific way


mat-1 (changes requested):
I don't think the struct name should be parsed and used for this since it has a bunch of [special cases](https://github.com/azalea-rs/azalea/blob/main/codegen/lib/code/blocks.py#L184) and is only meant for humans.
The `property.name` field is the one that's chosen by Minecraft and used for things like setblock commands.
Previously (like before [the commit I just made a minute ago](https://github.com/azalea-rs/azalea/commit/22baedb24b0d6b2b9148b3e0050d9bb36a4a54ab)) the names weren't guaranteed to be unique per-block (so brewing stands had has_bottle thrice) but now they should actually be suitable for this.

urisinger:
ill work on that soon, i think there should also be a way to set properies in a generic way, a map might not be the optimal solution, a set_property and get_property could also work

PR #148, opened 2024-05-14
Update simdnbt dep

veronoicc:
Just update simdnb git dep to make using the git dependency work again


mat-1 (approved):
thank you, lgtm <3

PR #149, opened 2024-05-22
make getting shapes faster

urisinger:
insted of a big match statment getting a shape is a static map now


veronoicc (approved):
I like

mat-1 (approved):
i made a commit so shape() falls back to SHAPE1 just in case we somehow have an invalid block id so this doesn't end up being a breaking change, lgtm now :)

PR #153, opened 2024-05-29
allow for ecs only clients

urisinger:
sorry for reoprning the pr i changed the branch name. 
 
 havent found an easy way to make both types of clients aviable, its possible that sync client might need to be a seprate crate or be behind a feature gate.


mat-1:
This exists now, as of https://github.com/azalea-rs/azalea/commit/8045b4eda284e58b40cb28b1528430e2d0b9d268

Some changes still need to be made for good auto-reconnecting though.

PR #155, opened 2024-06-16
Fix command packets

EnderKill98:
I finally tracked down, why commands don't work and cause the client to disconnect.

The packet has its contents copied from the "Signed" counterpart, which is unexpected. Thus most servers kick you for it.

This should address it.

Fixes https://github.com/azalea-rs/azalea-viaversion/issues/4


mat-1 (approved):
This looks right, thanks :)

PR #156, opened 2024-06-17
Added Left Click Mine Functionality

AS1100K:
## What this PR does?
This PR, will introduce a function in `azalea::Client` i.e. `left_click_mine`, once it is enabled it will make the bot mine the blocks it is looking at if it is within it's range In simpler words, it is a simple left-click implementation for mining blocks.

## Why this was added?
Although, it is quite basic functionality which could be used either for digging tunnels, mining stones in stone generator, etc. But I thought having it out of the box supported by azalea would be great, that's why I opened this PR.

## Why this PR was not opened on the main branch of azalea?
The main branch currently, supports Minecraft version `1.21` and introduces changes in `azalea-client/interact.rs` which interferes with `left_click_mine`. This [issue](https://discord.com/channels/1041923016856191016/1041923017317548064/1252128882971246612) makes `left_click_mine` not work in `1.21`. But I am working on a fix, and will soon open a PR to the main branch.

### DIFF
--- a/azalea-client/src/mining.rs
+++ b/azalea-client/src/mining.rs
@@ -1,6 +1,6 @@
 use azalea_block::{Block, BlockState, FluidState};
 use azalea_core::{direction::Direction, game_type::GameMode, position::BlockPos, tick::GameTick};
-use azalea_entity::{mining::get_mine_progress, FluidOnEyes, Physics, Position, LocalEntity};
+use azalea_entity::{mining::get_mine_progress, FluidOnEyes, Physics, LocalEntity};
 use azalea_inventory::ItemSlot;
 use azalea_physics::PhysicsSet;
 use azalea_protocol::packets::game::serverbound_player_action_packet::{
@@ -98,7 +98,6 @@ fn handle_auto_mine(
     mut query: Query<
         (
             &HitResultComponent,
-            &Position,
             Entity,
             Option<&Mining>,
             &InventoryComponent,
@@ -111,7 +110,6 @@ fn handle_auto_mine(
 ) {
     for (
         hit_result_component,
-        position,
         entity,
         mining,
         inventory,
@@ -121,14 +119,13 @@ fn handle_auto_mine(
     {
         let block_pos = hit_result_component.block_pos;
 
-        if (mining.is_none()
+        if mining.is_none()
             || !is_same_mining_target(
             block_pos,
             inventory,
             current_mining_pos,
             current_mining_item,
-        ))
-            && position.distance_to(&block_pos.to_vec3_floored()) <= 7.0
+        )
         {
             start_mining_block_event_writer.send(StartMiningBlockEvent {
                 entity,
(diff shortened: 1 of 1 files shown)


CuriousCodingCanadian:
lgtm, even though I have no permissions to say that

azalea-client/src/mining.rs
```diff
@@ -66,6 +76,66 @@ impl Client {
...
+        let block_pos = hit_result_component.block_pos;
+
+        if (mining.is_none()
+            || !is_same_mining_target(
+            block_pos,
+            inventory,
+            current_mining_pos,
+            current_mining_item,
+        ))
+            && position.distance_to(&block_pos.to_vec3_floored()) <= 7.0
```
mat-1, line 131: I think this line is unnecessary as `update_hit_result_component` already has its own reach check.

azalea-client/src/mining.rs
```diff
@@ -66,6 +76,66 @@ impl Client {
...
+        {
+            start_mining_block_event_writer.send(StartMiningBlockEvent {
```
mat-1, line 133: This should probably check `hit_result.miss` and send `StopMiningBlockEvent` in that case instead.

azalea-client/src/mining.rs
```diff
@@ -66,6 +76,66 @@ impl Client {
...
+        if enabled {
+            entity_mut.insert(AutoMine);
+        } else {
+            entity_mut.remove::<AutoMine>();
+        }
+    }
+}
+
+#[derive(Component)]
+pub struct AutoMine;
```
mat-1, line 95: Tbh I'm not a fan of this name since to me it sounds like it would automatically mine like Baritone. I was thinking like LeftClickHeld or something but that might be too verbose. Feel free to suggest better names, please :)
(maybe look at the decompiled Minecraft source for naming ideas, I do that sometimes)

mat-1 (changes requested):

azalea-client/src/mining.rs
AS1100K: I was thinking of calling it `LeftClickMine` as it is a implementation of holding left click but only for mining, it can't attack entities so `LeftClickHeld` might be little misleading. Other name i came up with was `TriggerLeftClick` or `TriggerLeftClickMine` but this might be too verbose. What do you think?

mat-1: Sure, `LeftClickMine` sounds fine.

AS1100K:
Should I also open this PR on main branch. Currently the error `AlreadyPresentError` is thrown but I think this has nothing to do with `LeftClickMine`.

mat-1:
> Should I also open this PR on main branch.
Yeah, I forgot that this PR isn't targeted at the main branch, I'd appreciate that.

> Currently the error `AlreadyPresentError` is thrown but I think this has nothing to do with `LeftClickMine`.
Wdym? Your branch and the main branch both work fine for me (at least when running the testbot example). What's your code that triggers that error?

AS1100K:
> Yeah, I forgot that this PR isn't targeted at the main branch, I'd appreciate that.

I have opened a PR #168 on the main branch.

mat-1 (approved):

Issue #130, opened 2024-03-02
auth customisation options for use outside of Azalea


1zun4:
I am currently looking forward to using azalea-auth for https://github.com/CCBlueX/LiquidLauncher to replace the horrible "miners" library, which has no proper error handling and keeps throwing error codes that a normal user cannot really interpret.

But there is one big problem, and that is that I cannot use my own client ID without forking this repository, and I do not want to use the Nintendo Switch client ID. Also, the fact that you cannot pass your own cache to auth, only a path to an existing cache file, is another thing that would require me to fork it. 

Since azalea is more of a library, it makes sense to give the programmer using the library more control without having to fork the repository and change the code.

PR #159, opened 2024-06-19
Auth Customization Options

AS1100K:
## What this PR adds?
This PR addresses and closes the issue #130. It introduces two new functions in `azalea_client::Account`:
* `microsoft_with_custom_client_id_and_scope` 
* `with_microsoft_access_token_and_custom_client_id_and_scope`.

This PR will allow the developer, to customize the `client_id` and `scope`, or use the nintendo switch one. Also, some other functions in `auth.rs` like `refresh_ms_auth_token`, `get_ms_link_code`, `interactive_get_ms_auth_token`, etc. are modified and takes some extra parameters now. See the changes in [`auth.rs`](https://github.com/azalea-rs/azalea/pull/159/files#diff-95c725220d95ac22f437570c310f8c41116efd3f4a1876f95cd38455f3bcf63f) for more information.

## Backwards Compatibility?
The existing functions `microsoft` and `with_microsoft_access_token` will maintain their current functionality. However, their implementations have been refactored to depend on the new functions `microsoft_with_custom_client_id_and_scope` and `with_microsoft_access_token_and_custom_client_id_and_scope` to avoid code duplication and facilitate easier maintenance.

### DIFF
--- a/azalea-auth/src/auth.rs
+++ b/azalea-auth/src/auth.rs
@@ -13,7 +13,7 @@ use thiserror::Error;
 use uuid::Uuid;
 
 #[derive(Default)]
-pub struct AuthOpts {
+pub struct AuthOpts<'a> {
     /// Whether we should check if the user actually owns the game. This will
     /// fail if the user has Xbox Game Pass! Note that this isn't really
     /// necessary, since getting the user profile will check this anyways.
@@ -25,9 +25,9 @@ pub struct AuthOpts {
     /// done.
     pub cache_file: Option<PathBuf>,
     /// If you choose to use your own microsoft authentication instead of using nintendo switch, just put your client_id here
-    pub client_id: Option<&'static str>,
+    pub client_id: Option<&'a str>,
     /// If you want to use custom scope instead of default one
-    pub scope: Option<&'static str>
+    pub scope: Option<&'a str>
 }
 
 #[derive(Debug, Error)]
@@ -63,7 +63,7 @@ pub enum AuthError {
 /// If you want to use your own code to cache or show the auth code to the user
 /// in a different way, use [`get_ms_link_code`], [`get_ms_auth_token`],
 /// [`get_minecraft_token`] and [`get_profile`] instead.
-pub async fn auth(email: &str, opts: AuthOpts) -> Result<AuthResult, AuthError> {
+pub async fn auth<'a>(email: &str, opts: AuthOpts<'a>) -> Result<AuthResult, AuthError> {
     let cached_account = if let Some(cache_file) = &opts.cache_file {
         cache::get_account_in_cache(cache_file, email).await
     } else {
--- a/azalea-auth/src/auth.rs
+++ b/azalea-auth/src/auth.rs
@@ -24,10 +24,12 @@ pub struct AuthOpts<'a> {
     /// The directory to store the cache in. If this is not set, caching is not
     /// done.
     pub cache_file: Option<PathBuf>,
-    /// If you choose to use your own Microsoft authentication instead of using Nintendo Switch, just put your client_id here.
+    /// If you choose to use your own Microsoft authentication instead of using
+    /// Nintendo Switch, just put your client_id here.
     pub client_id: Option<&'a str>,
-    /// If you want to use custom scope instead of default one, just put your scope here.
-    pub scope: Option<&'a str>
+    /// If you want to use custom scope instead of default one, just put your
+    /// scope here.
+    pub scope: Option<&'a str>,
 }
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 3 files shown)


1zun4:
This is great, but you are missing two important flags:
- RpsTicket format (see https://github.com/CCBlueX/azalea/commit/ca3e419904fd37b6f379c78242ca1e2b0139d9b5)
- Scope (see https://github.com/CCBlueX/azalea/commit/2b7484af0dd7bde6d68a7395f59d41e98ba7f2dc)

If you are using your own ClientID, these changes are also mandatory. Otherwise this option is useless.

AS1100K:
> I saw that you added a custom scope, but then removed it as it was apparently not needed? What is the reason for this? - [d2bb6f4](https://github.com/azalea-rs/azalea/commit/d2bb6f4b6d1350b5973cde7c0eeed2570c0dfe7f)

I removed the custom scope because that time i thought, that you will not require additional permissions but looking it again, i realize that having custom scope would allow to do more stuff, maybe send a email, etc.

I will commit add this back also RpsTicket format and commit the changes

veronoicc:
I just looked trough it an saw a lot of parameters with &'static str, and to my knowledge getting somethink as an &'static is not really possible (unless with a hardcoded str in this case) without leasing memory. Maybe using a non 'static str here is better?

veronoicc:
Meant fields in the AuthOps struct mb

azalea-client/src/account.rs
```diff
@@ -128,24 +137,35 @@ impl Account {
     /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
     /// let client = reqwest::Client::new();
     ///
-    /// let res = azalea_auth::get_ms_link_code(&client).await?;
+    /// let res = azalea_auth::get_ms_link_code(&client, "client_id", "scope").await?;
+    /// // Or, `azalea_auth::get_ms_link_code(&client, Some(client_id)).await?`
```
mat-1, line 141: The second line here should be taking two arguments.
```suggestion
    /// let res = azalea_auth::get_ms_link_code(&client, None, None).await?;
    /// // Or, `azalea_auth::get_ms_link_code(&client, Some(client_id), None).await?`
```

azalea-auth/src/auth.rs
```diff
@@ -280,25 +288,27 @@ pub enum GetMicrosoftAuthTokenError {
...
+/// let msa = azalea_auth::get_ms_auth_token(client, res, "client_id").await?;
 /// let minecraft = azalea_auth::get_minecraft_token(client, &msa.data.access_token).await?;
 /// let profile = azalea_auth::get_profile(&client, &minecraft.minecraft_access_token).await?;
 /// # Ok(())
 /// # }
 /// ```
 pub async fn get_ms_link_code(
     client: &reqwest::Client,
+    client_id: &str,
+    scope: &str
```
mat-1, line 305: I think this should take `Option<&str>`s to make it like some of the other functions and so the user isn't required to pass a client id and scope.
```suggestion
    client_id: Option<&str>,
    scope: Option<&str>
```

azalea-auth/src/auth.rs
```diff
@@ -314,19 +324,20 @@ pub async fn get_ms_link_code(
 pub async fn get_ms_auth_token(
     client: &reqwest::Client,
     res: DeviceCodeResponse,
+    client_id: &str
```
mat-1, line 327:
```suggestion
    client_id: Option<&str>
```

azalea-auth/src/auth.rs
```diff
@@ -357,15 +372,17 @@ pub async fn get_ms_auth_token(
 pub async fn interactive_get_ms_auth_token(
     client: &reqwest::Client,
     email: &str,
+    client_id: &str,
+    scope: &str
```
mat-1, line 376:
```suggestion
    client_id: Option<&str>,
    scope: Option<&str>
```

azalea-auth/src/auth.rs
```diff
@@ -280,25 +288,27 @@ pub enum GetMicrosoftAuthTokenError {
...
 /// ```
 /// # async fn example(client: &reqwest::Client) -> Result<(), Box<dyn std::error::Error>> {
-/// let res = azalea_auth::get_ms_link_code(&client).await?;
+/// let res = azalea_auth::get_ms_link_code(&client, "client_id", "scope").await?;
 /// println!(
 ///     "Go to {} and enter the code {}",
 ///     res.verification_uri, res.user_code
 /// );
-/// let msa = azalea_auth::get_ms_auth_token(client, res).await?;
+/// let msa = azalea_auth::get_ms_auth_token(client, res, "client_id").await?;
```
mat-1, line 296: These "client_id" and "scope" should just be `None` so the example works by default.

azalea-auth/src/auth.rs
```diff
@@ -335,6 +346,10 @@ pub async fn get_ms_auth_token(
             .json::<AccessTokenResponse>()
             .await
         {
+            if client_id != CLIENT_ID {
+                access_token_response.access_token.insert_str(0, "d=");
+            }
+
```
mat-1, line 352: There's other client ids where `d=` shouldn't be added, I think the user should be responsible for adding it themselves (get_minecraft_token already documents it).
```suggestion
```

azalea-auth/src/auth.rs
```diff
@@ -24,6 +24,10 @@ pub struct AuthOpts {
     /// The directory to store the cache in. If this is not set, caching is not
     /// done.
     pub cache_file: Option<PathBuf>,
+    /// If you choose to use your own microsoft authentication instead of using nintendo switch, just put your client_id here
```
mat-1, line 27: Tiny nitpick here but I've been trying to keep the grammar in the docs consistent, mainly just capitalizing proper nouns (so Microsoft and Nintendo Switch should be capitalized here) and adding periods at the end of sentences.

azalea-client/src/account.rs
```diff
@@ -128,24 +137,35 @@ impl Account {
...
+    /// // if you want to use your own client_id
     /// println!(
     ///     "Go to {} and enter the code {}",
     ///     res.verification_uri, res.user_code
     /// );
-    /// let msa = azalea_auth::get_ms_auth_token(&client, res).await?;
+    /// let msa = azalea_auth::get_ms_auth_token(&client, res, "client_id").await?;
```
mat-1, line 147:
```suggestion
    /// let msa = azalea_auth::get_ms_auth_token(&client, res, None).await?;
```

mat-1 (changes requested):
Thank you for making this, I'm sorry for taking so long to get to it. This looks good, the main thing I think should be changed is using `Option<&str>`  instead of `&str` for all the public functions that take a client_id and scope. I can resolve everything here tomorrow (I'm going to sleep after writing this) unless you want to.

By the way, I recommend you also run `cargo fmt` to standardize the formatting on some of the code.

mat-1 (approved):

PR #163, opened 2024-07-12
Determine rust channel by parsing rustc output if env vars do not exist

itsCryne:
The RUSTUP_TOOLCHAIN environment variable might not always be present. This is the case for e.g. NixOS where rust is routinely not installed via rustup, thus not setting this env var, causing build failures. Instead, build.rs will now run `rustc -V` and check if the output contains the word "nightly".

### DIFF
--- a/azalea/build.rs
+++ b/azalea/build.rs
@@ -1,7 +1,11 @@
+use std::env;
 use std::process::Command;
 
 fn main() {
-    let rustc_version_output = Command::new("rustc").arg("-V").output().unwrap();
+    let rustc_command = env::var("RUSTC")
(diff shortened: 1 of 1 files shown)


EightFactorial:
The copy of `rustc` in your path, if there even is one, isn't necessarily the version being used to build the project.

Maybe use the [`RUSTC`](https://doc.rust-lang.org/cargo/reference/config.html#buildrustc) environment variable?

itsCryne:
This will now also give an error if the beta channel is used

itsCryne:
Is there anything left to do before this can get reviewed/merged?

mat-1 (approved):
Thank you :)

PR #164, opened 2024-07-15
Emit configuration packets

ShayBox:


azalea-client/src/events.rs
```diff
@@ -82,7 +95,7 @@ pub enum Event {
     /// # }
     /// # }
     /// ```
-    Packet(Arc<ClientboundGamePacket>),
+    Packet(ClientboundPacket),
```
mat-1, line 98: This would require you to have to start writing `Event::Packet(ClientboundPacket::Game(packet)) => ...` in the match statements which is kinda too verbose imo. Maybe it should be split into `Event::GamePacket` and `Event::ConfigurationPacket` instead.

mat-1 (changes requested):
One minor nitpick but other than that this looks good :)

azalea-client/src/events.rs
ShayBox: I didn't know you could even do that, I've been doing match.

PR #167, opened 2024-07-20
Made `Hunger` and `DataComponentPatch` public

### LINKED ISSUE #166: `Hunger` and `DataComponentPatch` component are not public
Well `Hunger` is an important component and it is only accessible by `Client` i.e. in ecs environment it can't be used.

This is also similar with `DataComponentPatch` as it holds a lot of information about the item.

I have fixed these in my [fork](https://github.com/as1100k-forks/azalea) but azalea should have these components public by default.

_I came across these component, there can be more components which aren't public_

AS1100K:
This PR fixes #166 by making these components/structs public.

### DIFF
--- a/azalea-client/src/lib.rs
+++ b/azalea-client/src/lib.rs
@@ -36,7 +36,7 @@ pub use client::{
     TickBroadcast,
 };
 pub use events::Event;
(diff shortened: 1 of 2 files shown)


mat-1 (approved):
Thanks!

PR #168, opened 2024-07-20
Added Left Click Mine (1.21)

AS1100K:
This PR is similar to #156 but instead this is on main branch.

## What this PR does?
This PR, will introduce a function in `azalea::Client` i.e. `left_click_mine`, once it is enabled it will make the bot mine the blocks it is looking at if it is within it's range In simpler words, it is a simple left-click implementation for mining blocks.

## Why this was added?
Although, it is quite basic functionality which could be used either for digging tunnels, mining stones in stone generator, etc. But I thought having it out of the box supported by azalea would be great, that's why I opened this PR.


mat-1 (approved):
The `LeftClickMine` struct should probably have docs, but not a blocker for merging, lgtm otherwise ¯\\\_(ツ)\_/¯

PR #171, opened 2024-09-28
1.21.2

mat-1:


mat-1:
Major missing features:
- Sending ServerboundTickEndPacket and ServerboundPlayerInputPacket
- Updated block data (right now it gets most block ids - and thus shapes - wrong since they're shifted)

PR #174, opened 2024-10-21
Impl FromStr for BlockPos & Vec

ShayBox:
This can parse `X Y Z` back into a `BlockPos` or `Vec3` from the existing `Display` trait.


mat-1 (changes requested):
This looks good, though, could you write doc comments for the from_str implementations please (just a short sentence for each is fine)

mat-1 (approved):
Thank you :)

PR #176, opened 2024-10-21
Fix chat duplication from nested siblings

ShayBox:
This seems to fix the existing chat duplication bug from duplicate nested siblings.


mat-1:
Fixed by https://github.com/azalea-rs/azalea/commit/63fef9d94fbdafdf6beec408b733915dce2c0be7

PR #179, opened 2024-10-29
feat(registry): Derive Serialize and Deserialize on registry types

GaspardCulis:
Derives serde's `Serialize` and `Deserialize` on registry types defined using the `registry` macro when the `serde` feature is enabled.


mat-1 (approved):

mat-1:
thanks :)

PR #188, opened 2024-11-24
Feat: Host Documentation for multiple versions/branch

AS1100K:
I love azalea, and sometimes I need to use different branch of azalea like `1.20.4`. The issue is that azalea, currently only hosts documentation for `main` branch and though it would be great if it supported documentation multiple branches. This is what this PR introduces.

A Demo is available at https://as1100k.github.io/azalea/ [Old due to changes requested]

## Prerequisite
1. Create a [orphan branch](https://stackoverflow.com/questions/1384325/in-git-is-there-a-simple-way-of-introducing-an-unrelated-branch-to-a-repository#:~:text=git%20checkout%20%2D%2Dorphan%20newbranch%0Agit%20rm%20%2Drf%20.%0A%3Cdo%20work%3E%0Agit%20add%20your%20files%0Agit%20commit%20%2Dm%20%27Initial%20commit%27) `docs`. And a `Initial Commit` with `.gitignore` with the following content:
    ```
    target
    ```
2. Update `doc.yml` for old branches. _(If you approve, I will open PR for all those other branches)_
3. Change settings to host GitHub Pages from `docs` branch instead of `GitHub Actions`

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -41,18 +41,18 @@ jobs:
           rm -rf versioned_docs  # Clean up
           rm -rf target # Clean up
 
-      - name: Generate README.md
+      - name: Generate BRANCHES.md
         run: |
           VERSIONS=$(ls -d */ | sed 's#/##' | sort -r)  # Get all version directories and sort them (latest first)
           rm README.md # Delete README.md and generate a new one
-          echo "# Azalea Docs" > README.md
-          echo "Welcome to the documentation for Azalea Crate." >> README.md
-          echo "" >> README.md
+          echo "# Azalea Docs" > BRANCHES.md
+          echo "Welcome to the documentation for Azalea Crate." >> BRANCHES.md
+          echo "" >> BRANCHES.md
           
           # Update README.md with available versions
-          echo "## Available Versions" >> README.md
+          echo "## Available Versions" >> BRANCHES.md
           for VERSION in $VERSIONS; do
-            echo "- [$VERSION](./$VERSION/index.html)" >> README.md
+            echo "- [$VERSION](https://azalea.matdoes.dev/$VERSION/index.html)" >> BRANCHES.md
           done
 
           echo "README.md updated successfully."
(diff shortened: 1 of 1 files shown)


ShayBox:
I've always viewed old docs on docs.rs for the published last version, or local cargo docs if I have to modify an old version, but I also prefer to use zenith proxy or via proxy to connect to old versions with the latest Azalea

AS1100K:
That's correct, but the whole point of this PR is convenience for viewing docs for multiple branches. Also, it's not always that the branch is same with the docs on docs.rs as 1.20.6 branch is ahead by a commit<https://github.com/azalea-rs/azalea/tree/1.20.6>.
Also, azalea's scope is to be up-to date with the latest version of minecraft and keeping older versions updated are out of scope as using proxy like ViaProxy or ZenithProxy works. But there are scenario where using proxy doesn't turns out to be effective, as in my case ZenithProxy was limiting the bot from mining additional layer of blocks because of latency or lag and since, i was hosting on a free server, I didn't want to go crazy hosting another instance just for ZenithProxy.
Although you are right, we can access docs via docs.rs or using cargo doc locally but I though it would be great hosting those all on a single website.
________________________________

mat-1:
Oh also I just realized, it should have a filter to only do the commit on the docs branch if the current branch looks like a version number. (since I'll often have temporary branches for other stuff)

AS1100K:
Listings are now on `branches.html` [Demo](https://as1100k.github.io/azalea/branches) and opening azalea.matdoes.dev redirects to the main branch documentation now

mat-1 (approved):

mat-1:
I just merged this now since I didn't want to leave it open for too long but after merging I noticed a few more issues (it breaks old links to docs, adds noise to the Discord webhook).. those issues are solvable with some effort but tbh after considering this more I don't think I actually want this in Azalea since it's extra maintenance burden for a "feature" that's barely maintained and not entirely supported anyways. I think people should just generate the docs on their local machines if they want them for a branch. Sorry.

AS1100K:
Thanks for trying it out. I do agree that it will add extra maintenance burden on azalea, since it's goal is to keep up to date with latest minecraft version, not support multiple versions.

PR #189, opened 2024-11-26
1.21.4

mat-1:


mat-1:
not tested in game yet, and everything from here still applies: https://github.com/azalea-rs/azalea/pull/171#issuecomment-2430915354

mat-1:
the `refactor-azalea-protocol` branch is based on this pr and changes a lot, so... non-trivial updates are going there instead (it's going to be merged immediately after this one is merged anyways so the distinction isn't exactly significant)

PR #190, opened 2024-11-27
Refactor azalea-protocol

### LINKED ISSUE #173: Whisper message is wrong
When printing a whisper chat message from some player to the client (via the `message` fn), it prints:
`You whisper to {player}: {message}`

It should be:
`{player} whispers to you: {message}`

For a minimal code example that displays this, just use the example from the README:
```rust
//...

#[derive(Default, Clone, Component)]
pub struct State {}

async fn handle(bot: Client, event: Event, state: State) -> anyhow::Result<()> {
    match event {
        Event::Chat(m) => {
            println!("{}", m.message().to_ansi());
        }
        _ => {}
    }

    Ok(())
}
```

mat-1:
Fixes some things in azalea-protocol that have been bothering me for a while:
- Significantly simplify usage of the `declare_state_packets` macro (notably, packet ids are implicit now)
- Rename packet modules from like `serverbound_keep_alive_packet` to `s_keep_alive`
- Rename `configuration` and `handshaking` to `config` and `handshake`
- Remove the "Packet" suffix for packet structs (since the Clientbound/Serverbound prefix already indicates that it's a packet)
- Re-export all packet structs so they're accessible through like `azalea_protocol::game::ClientboundKeepAlive`
- Rename `ItemSlot` to `ItemStack` for consistency with Mojmap
- Replace `.get()` with `.into_variant()` from a new `Packet` trait, which is used to avoid most uses of that function now. 
- Rename `read_from`/`write_into` to `azalea_read`/`azalea_write` and `McBufReadable`/`McBufWritable` to `AzaleaRead`/`AzaleaWrite`
- Rename `McBuf` to `AzBuf`
- Add an attribute `#[limit(number)]` so strings can be limited to a certain size.

Based on [the 1.21.4 PR](https://github.com/azalea-rs/azalea/pull/189) since it's probably going to be merged in a few days anyways ([diff](https://github.com/azalea-rs/azalea/compare/a2a941f92aa33a1c5b198206d4d061c8b681c86a..refactor-azalea-protocol)). Also, these changes were mostly done with a lot of find-and-replace, so I may have messed up in some places without realizing. Please let me know if you notice something I broke or if you disagree with a change.

Some changes were inspired by @EightFactorial's [FrogLight](https://github.com/EightFactorial/FrogLight) project <3

Closes #173


mat-1:
I realized I should probably merge this sooner rather than later since it fixes a lot, so I downgraded it to 1.21.3 and merged it for that instead of waiting for 1.21.4 to release.

PR #195, opened 2024-12-10
Disable the `deadlock_detection` feature by default

EightFactorial:
Fixes conflicts with any packages that enable `parking_lot`'s `send_guard` feature


mat-1 (dismissed):
This is a little unfortunate but makes sense, I'll change the testbot example to also mention that the feature has to be enabled for deadlock detection to work.

mat-1 (approved):

PR #196, opened 2024-12-11
Implement EntityPositionSync

mat-1:
Minecraft now has a `vec_delta_codec` field for entities which has something to do with fixing position desyncs for entities. There's also now a `ClientboundEntityPositionSync` packet that updates the position of an entity to an exact value (instead of a delta), unless it's called on a player in which case it only updates the vec_delta_codec.

This PR implements those, which means it'll hopefully fix entity position desync issues if you were having those. I did test it in-game, though not extensively, which is why I'm leaving this PR open for a bit just in case there's issues found with it.


ShayBox:
It still de-syncs entity positions, but if you jump the sync packet fixes it until you send another regular movement packet again, then it goes back to the de-synced position.

https://youtu.be/ZRG-3FbAaY8

```rust
use azalea::{
    app::{App, Plugin},
    ecs::prelude::*,
    entity::{metadata::Player, EyeHeight, LocalEntity, Position},
    nearest_entity::EntityFinder,
    physics::PhysicsSet,
    prelude::*,
    LookAtEvent,
};

/// Automatically stare at the closest player
pub struct AutoLookPlugin;

impl Plugin for AutoLookPlugin {
    fn build(&self, app: &mut App) {
        app.add_systems(GameTick, handle_auto_look.before(PhysicsSet));
    }
}

pub fn handle_auto_look(
    mut query: Query<Entity, (With<LocalEntity>, With<Player>)>,
    entities: EntityFinder<With<Player>>,
    targets: Query<(&Position, Option<&EyeHeight>)>,
    mut look_at_events: EventWriter<LookAtEvent>,
) {
    for entity in &mut query {
        let Some(target) = entities.nearest_to_entity(entity, f64::MAX) else {
            continue;
        };

        let Ok((target_pos, target_eye_height)) = targets.get(target) else {
            continue;
        };

        let mut position = **target_pos;
        if let Some(eye_height) = target_eye_height {
            position.y += f64::from(**eye_height);
        }

        look_at_events.send(LookAtEvent { entity, position });
    }
}
```

PR #197, opened 2024-12-12
chore(core): Derive Serialize and Deserialize for `direction` types

GaspardCulis:
Derive serde's Serialize and Deserialize for `direction` types when crate's `serde` feature is enabled.

Not a big change but is a continuation of #183 , #179 and #175


mat-1 (approved):
sure, lgtm

PR #208, opened 2025-03-10
Implement a to_html method for FormattedText

Kumpelinus:
Also fix a small issue with ansi formatting where it duplicated text.

### DIFF
--- a/azalea-chat/src/component.rs
+++ b/azalea-chat/src/component.rs
@@ -67,6 +67,44 @@ impl FormattedText {
         }
     }
 
+    /// Render all components into a single `String`, using your custom
+    /// closures to drive styling, text transformation, and final cleanup.
+    ///
+    /// # Type params
+    /// - `F`: `(running, component, default) -> (prefix, suffix)` for per-component styling
+    /// - `S`: `&str -> String` for text tweaks (escaping, mapping, etc.)
+    /// - `C`: `&final_running_style -> String` for any trailing cleanup
+    ///
+    /// # Args
+    /// - `style_formatter`: how to open/close each component’s style
+    /// - `text_formatter`:  how to turn raw text into output text
+    /// - `cleanup_formatter`: emit after all components (e.g. reset codes)
+    /// - `default_style`:    where to reset when a component’s `reset` is true
+    ///
+    /// # Example
+    /// ```rust
+    /// use azalea_chat::{FormattedText, DEFAULT_STYLE};
+    /// use serde::de::Deserialize;
+    ///
+    /// let component = FormattedText::deserialize(&serde_json::json!({
+    ///    "text": "Hello, world!",
+    ///    "color": "red",
+    /// })).unwrap();
+    ///
+    /// let ansi = component.to_custom_format(
+    ///     |running, new, default| (running.compare_ansi(new, default), String::new()),
+    ///     |text| text.to_string(),
+    ///     |style| {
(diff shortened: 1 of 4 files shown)


Kumpelinus:
Maybe we should use classes instead of doing the formatting directly (for everything besides color)

Kumpelinus:
Also I will add tests before a merge.

azalea-chat/src/component.rs
```diff
@@ -89,41 +135,26 @@ impl FormattedText {
...
+            &DEFAULT_STYLE,
+        )
     }
 
-    /// Convert this component into an
-    /// [ANSI string](https://en.wikipedia.org/wiki/ANSI_escape_code).
-    ///
-    /// This is the same as [`FormattedText::to_ansi`], but you can specify a
-    /// default [`Style`] to use.
-    pub fn to_ansi_with_custom_style(&self, default_style: &Style) -> String {
```
mat-1, line 101: I'd like to keep `to_ansi_with_custom_style` to avoid breaking changes and so if the user wants a custom default style then they don't have to copy the code from `to_ansi`.

Btw, the reason this exists at all is to make it easier if the user is doing something like logging all of their text in a certain color and they want to make sure the text stays that color when `reset` gets set to true. This doesn't matter for HTML since in HTML you can just make a parent element with the default style.

azalea-chat/src/component.rs
```diff
@@ -89,41 +135,26 @@ impl FormattedText {
...
-
-            running_style.apply(component_style);
-        }
-
-        if !running_style.is_empty() {
-            built_string.push_str("\u{1b}[m");
-        }
-
-        built_string
+    pub fn to_html(&self) -> String {
```
mat-1, line 146: There should probably be at least one test for this (see the to_ansi tests in `text_component.rs` and `integration_test.rs`)

azalea-chat/src/component.rs
```diff
@@ -89,41 +135,26 @@ impl FormattedText {
...
+        self.to_custom_format(
+            |_, new, _| {
+                (
+                    format!("<span style=\"{}\">", new.get_html_style()),
```
mat-1, line 150: This should be `running.merged_with(new).get_html_style()`, otherwise it's wrong when a style only partially changes and the previous style needs to be preserved.

For example, `§l§aA§bB` was getting turned into `<span style="color: #55FF55;font-weight: bold;">A</span><span style="color: #55FFFF;">B</span>` when it should be `<span style="color: #55FF55;font-weight: bold;">A</span><span style="color: #55FFFF;font-weight: bold;">B</span>`

azalea-chat/src/component.rs
```diff
@@ -89,41 +135,26 @@ impl FormattedText {
...
+                    "</span>".to_owned(),
+                )
+            },
+            |text| text.replace("\n", "<br>"),
```
mat-1, line 154: This should escape the HTML (adding `.replace("<", "&lt;").replace(">", "&gt;")` should be enough)

mat-1 (changes requested):
I like the way you implemented this, the function arguments in to_custom_format seem sensible to me. I don't see any major issues in the way this was designed.

Kumpelinus:
My bad. Not done yet

Kumpelinus:
Now it should be (unless you need more tests ofc. I will add them in that case).

mat-1 (approved):
Thanks!

PR #209, opened 2025-03-12
update dead wiki.vg links

Gjum:
Since wiki.vg is now [offline](https://tkte.ch/articles/2024/11/11/sunsetting.html), I've replaced all links with their counterparts on `https://minecraft.wiki/w/Minecraft_Wiki:Projects/wiki.vg_merge`


mat-1 (approved):
Thanks :)

PR #210, opened 2025-03-16
Fluid physics fixes

mat-1:
- Fixes the bot incorrectly gaining upwards velocity in some cases in fluids
- Fixes the bot being influenced by fluids that are too far by one block in positive coordinate directions.
- Implements force_solid which resolves some problems related to water flowing

### DIFF
--- a/azalea-block/src/behavior.rs
+++ b/azalea-block/src/behavior.rs
@@ -56,7 +56,6 @@ impl BlockBehavior {
         self
     }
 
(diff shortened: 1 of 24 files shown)


mat-1:
Unfinished, there's still a bug where fluids that you're not touching can incorrectly push you. Also, there's a bunch of debug and commented-out code lying around that needs to be removed when this is ready.

PR #216, opened 2025-04-18
remove `.unwrap()` from `wait_until_goto_target_reached`

manen:
sometimes the tick broadcaster would lag and would cause the entire thread to panic with `Lagged(1)`


mat-1 (approved):
Thanks!

PR #220, opened 2025-04-26
add distance param for ReachBlockPosGoal

x-osc:
idk if what im doing is good or performant so

### DIFF
--- a/azalea/src/pathfinder/goals.rs
+++ b/azalea/src/pathfinder/goals.rs
@@ -194,6 +194,7 @@ impl<T: Goal> Goal for AndGoals<T> {
 #[derive(Clone, Debug)]
 pub struct ReachBlockPosGoal {
     pub pos: BlockPos,
+    pub distance: f32,
     pub chunk_storage: ChunkStorage,
 }
 impl Goal for ReachBlockPosGoal {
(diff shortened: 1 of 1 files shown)


mat-1 (changes requested):
Thanks for the PR :).

Most people probably won't be sure what the client's default pick range is, so I'd recommend writing a `new` impl for ReachBlockPosGoal that takes `(pos: BlockPos, chunk_storage: ChunkStorage)` as the arguments and uses the 4.5 by default range. There should also be a `new_with_distance(pos, chunk_storage, distance)` to go along with it (which is called by `new`).

Regarding performance, `Goal::success` is called very frequently while pathfinding so it generally is worth micro-optimizing. I think the best solution here would be to change `max_pick_range` and `actual_pick_range` to be precalculated - so just fields in ReachBlockPosGoal that are calculated during the `new_with_distance` call. Also, `max_pick_range` probably wasn't the best name on my part (since it's not used by the actual `pick` function) feel free to rename it to like `max_player_range` or something.

By the way, out of curiosity, what are you actually using this for?

x-osc:
thx for reviewing :D
should be done now

im using this for [https://github.com/x-osc/lickbot](https://github.com/x-osc/lickbot) 
right now im working on making it auto mine ores and stuff

mat-1 (approved):
LGTM, thanks!

PR #224, opened 2025-06-09
Add resource pack example

Kumpelinus:


mat-1:
Tbh, I'm not sure that this is important enough to warrant its own dedicated example here. The main thing this example does is highlight the existence of the `ResourcePackPush` packet, though I think the same thing could be accomplished by writing docs for that type to make it easier to find. Alternatively, maybe a bigger example could be written that does more things with resource packs (maybe something like disabling `AcceptResourcePacksPlugin` and downloading the pack and sending the right packets depending on the response)

mat-1:
Closing for now, feel free to re-open if you wanna continue this.

PR #229, opened 2025-07-11
Add TicksAlive component

Kumpelinus:


Kumpelinus:
Also I don't know what the best name for the Plugin / Component would be

mat-1 (changes requested):
Thanks for writing this, I don't have any major complaints. :)

I'll also mention here that it'd probably be a bit more efficient to just have a global tick counter resource and instead of incrementing the component for every entity, we just store the exact tick when we joined. However this would make the component harder to use and I don't think the performance impact would be significant, so I think the current approach is fine.

Also, nice test. <3

mat-1 (approved):
Thanks :)

PR #230, opened 2025-08-02
Add best effort for default max stack sizes

EnderKill98:
Maybe useful, was implementing a bunch of inventory things and noticed this info missing, which can cause issues with click simulation.

This is by no means 100%, but I'd like to think it's close to 100% complete.

Not pretty code, but should get the job done until some better way is decided on.


EnderKill98:
Seems clippy is not complaining about me

mat-1:
Obsoleted by #232

EnderKill98:
No worries. Mainly needed it, since I added some stuff to do item management and needed a way to know how much count was "free" in any given slot. Just saw the PR you made as well. Huge!

PR #231, opened 2025-08-02
Add FAQ to README

EightFactorial:


mat-1 (approved):
Thanks!

I was thinking this might fit better on the rustdoc documentation, but that already mentions logging so it's probably fine to also mention it here.

Also, it might be better to recommend a level other than `off` (like `error`) instead, buuut I don't think it matters too much.

PR #233, opened 2025-08-07
Improve System Chat Regex & Add System Whisper Regex

ShayBox:
This separates the system chat sender/content splitting regex and adds a system chat whisper regex which can both be overridden to add custom server chat format parsing, the built-in ones support Vanilla and Essentials Chat, and now use capture group names so that really weird formats with out of order groups can still be parsed (such as content, then sender)

### DIFF
--- a/azalea-client/src/plugins/chat/mod.rs
+++ b/azalea-client/src/plugins/chat/mod.rs
@@ -70,7 +70,7 @@ impl ChatPacket {
                 }
 
                 // It's a system message, so we'll have to match the content with regex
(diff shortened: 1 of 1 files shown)


mat-1 (changes requested):
Thanks, Shay :)

Tbh, I'm not really convinced that these global variables are a good idea. They make certain things more annoying (like unit testing), and I don't think there's much benefit to having them in this case (since the user can just write their own functions to parse messages if they really need to anyways).

Anyways, I do welcome the improvements to the regexes, since I'd like them to be reasonable defaults that users usually don't have to think about, so I would appreciate it if you could update it to just keep those improvements without the global variables.

ShayBox:
This should probably be squash merged into one commit instead of merging a force push.

mat-1 (approved):

PR #236, opened 2025-08-13
fix: mining fails in Creative Mode

qwqawawow:
sloppy code: bot.mine() is not working in Creative mode.（spigot&vanilla）
 works on vanilla (open to LAN) & paper with GrimAC


mat-1 (changes requested):
Thanks for the PR!

The two additional triggers look good to me, but I think the code you removed (the `**mine_delay = 5` and the stuff in `continue_mining_block`) should be kept since those are what handle the case where left-click is held down (since vanilla adds some extra delays when you do that).

eihqnh:
> Thanks for the PR!
> 
> The two additional triggers look good to me, but I think the code you removed (the `**mine_delay = 5` and the stuff in `continue_mining_block`) should be kept since those are what handle the case where left-click is held down (since vanilla adds some extra delays when you do that).

got it, thank you very much

Additional changes:
The packet sending order for creative mode in the `continue_mining_block` function seemed reversed, which affected `bot.left_click_mine`. It has been fixed and tested.

Thank you for your patient explanation and guidance :3

PR #237, opened 2025-08-13
Sneaking

### LINKED ISSUE #217: Sneaking/Crouching
No way to currently sneak using azalea. I think this is a major requirement, and should be added in the following updates.
Thanks.

mat-1:
Finally finishing up that old `sneaking` branch. Adds a new `azalea_entity::Sneaking` component for when we're trying to sneak. Tested with Grim and doesn't break any of the current tests.

Closes #217


mat-1:
(Changing the eye height hasn't been implemented yet)

PR #240, opened 2025-08-24
add property_map and get_property to block

urisinger:


mat-1 (approved):
Thanks :)

I was writing a review and as I got further into it I realized some of my proposed changes would be way easier if I did some refactoring...

I mostly split some things into separate functions/modules and improved some docs; more cleanup would be possible but I didn't find it necessary. Also, the codegen'd code now writes properties like `X = "x"` to avoid having to convert identifiers back to snake case.

The public changes in my commit are that properties now impl `Display` instead of `ToString` (so they can be in format macros), the `Property` trait now has a `to_static_str` function, the new functions on `BlockTrait` use `&'static str` instead of `String`, and they have docs.

PR #246, opened 2025-09-14
Fix nix dev env

Kumpelinus:


Kumpelinus:
Idk why it got opened here :skull:

Issue #251, opened 2025-09-24
inventory gets closed after it gets opened


Hiradpi:
if the server sends the client close_window and then open_window for some reason open window gets prossesed faster than close window

the issue is in the `handle_client_side_close_container_event` 
the open inventory packet for some reason gets prossesed faster than close_window 

Opened container with ID 1 
Client-side closed container with ID 1 

<img width="1145" height="152" alt="Image" src="https://github.com/user-attachments/assets/6f6a3e85-2a11-4648-8121-3f12a7006323" />

mat-1:
Added a test for this in https://github.com/azalea-rs/azalea/commit/dfeb0880eeb833ca6653fef197318fa985009cfb that previously failed and now passes.

PR #252, opened 2025-09-24
fix for inventory update issue

Hiradpi:
https://github.com/azalea-rs/azalea/issues/251

### DIFF
--- a/azalea-client/src/plugins/inventory.rs
+++ b/azalea-client/src/plugins/inventory.rs
@@ -801,8 +801,13 @@ pub fn handle_client_side_close_container_event(
 ) {
     for event in events.read() {
         let mut inventory = query.get_mut(event.entity).unwrap();
-
-        // copy the Player part of the container_menu to the inventory_menu
+        if inventory.id != event.window_id {
+            warn!(
+                "Got ClientSideCloseContainerEvent for container with ID {}, but the current container ID is {}",
+                event.window_id, inventory.id
+            );
+            continue;
+        }
         if let Some(inventory_menu) = inventory.container_menu.take() {
             // this isn't the same as what vanilla does. i believe vanilla synchronizes the
             // slots between inventoryMenu and containerMenu by just having the player slots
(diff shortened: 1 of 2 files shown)


mat-1:
Thanks for the PR :).

I wonder if this is the correct solution; have you checked if this is what Minecraft does? (if you don't have the decompiled code you can get it with something like [DecompilerMC](https://github.com/hube12/DecompilerMC)).

Based on the info in your issue, I would guess that the root cause of the problem is purely due to ECS ordering that I forgot to account for. With Bevy's ECS (in theory), each system function has to have an explicit ordering, which means that two events of different types that are sent at the same time will always run their corresponding systems in the same order, regardless of the order that they were sent in (as long as they were received in the same ECS schedule update).

Here, the events for close+open are being handled in the wrong order because the system that handles close events always happens after the one for open events. You could swap them, but then it'd cause problems for the reverse situation when we receive open+close in the same update.

My usual solution has been to replace the relevant event handling systems with Bevy observers/triggers, which are like normal systems except that they happen immediately when they're triggered. I would've used them in Azalea more, but they're a relatively new Bevy feature. Here's Bevy's docs for them: https://docs.rs/bevy/latest/bevy/prelude/struct.Observer.html . If you think this would be a good solution, you should search "trigger" in Azalea's code to find a bunch of examples (for example, ReceiveHelloEvent seems kinda similar since it's also for a packet handler). To avoid breaking changes, I also sometimes add a system for each event that reads the events with EventReader and triggers their observer function (see handle_swing_arm_event for an example).

Sorry if this is too much and that I wrote so much yap. Also sorry for writing the buggy code in the first place :3. Please lmk here if you need help.

Hiradpi:
well for now before you rewrite the handling systems this would be a hardfix im not sure if the vanilla mc uses the same logic but its working perfect for me

mat-1:
> well for now before you rewrite the handling systems this would be a hardfix im not sure if the vanilla mc uses the same logic but its working perfect for me

Looking at Minecraft's code for handling this packet, the window ID is indeed completely ignored :pensive: 
image

Your patch will work for some cases, but I imagine that there's some weird servers that rely on the client not checking the window ID. So, it's not a correct fix and I can't really merge it :(

PR #260, opened 2025-10-04
fix(inventory): update 1.21.9 entity and block entity component struc…

eihqnh:
## NOTE
This is a dirty fix for BlockEntityData component and small improvement for EntityData(?)
The code quality is not quite up to standard — consider this a small hint for improvement(?)
 hope useful and sorry for being sloppy with the code :3 

## Issue

In version 1.21.9, command blocks with special NBT data fail to parse because the ComponentType for block entities is TypedEntityData<BlockEntityData>, affecting command blocks and possibly other related blocks (not fully tested).

Similarly, TypedEntityData<EntityData> for entity types seems to affect spawner eggs (I haven’t tested this yet).

However, I have fully tested my changes, and they work as expected.

## Change
Default entity data for spawn eggs now uses `EntityKind` enums instead of raw NBT.
Codegen updated to map NBT `id` fields to `EntityKind` variants automatically.

## source ref (fabric Yarn)
- `net/minecraft/entity/TypedEntityData.java`
- `net/minecraft/component/DataComponentTypes.java`


mat-1 (changes requested):
Oh, thank you!

The TypedEntityData struct feels a bit awkward to me since it's not really doing anything useful... I think the code would be clearer if the structs were just like
```rs
pub struct EntityData {
    pub kind: EntityKind,
    pub data: NbtCompound,
}
```
and
```rs
pub struct BlockEntityData {
    pub kind: BlockEntityKind,
    pub data: NbtCompound,
}
```
instead. This would let the AzBuf derive macro work too ;3.

Also I wanna mention a couple things about style (not that they'd matter after my proposed changes tho): I typically use the word "kind" instead of "type" since `type` is a reserved keyword in Rust, and also I like having a space after the `//` in comments.

eihqnh:
> Oh, thank you!
> 
> The TypedEntityData struct feels a bit awkward to me since it's not really doing anything useful... I think the code would be clearer if the structs were just like
> 
> ```rust
> pub struct EntityData {
>     pub kind: EntityKind,
>     pub data: NbtCompound,
> }
> ```
> 
> and
> 
> ```rust
> pub struct BlockEntityData {
>     pub kind: BlockEntityKind,
>     pub data: NbtCompound,
> }
> ```
> 
> instead. This would let the AzBuf derive macro work too ;3. To make them serialize with serde correctly you might have to put like `#[serde(rename = "id")]` and `#[serde(flatten)]` for the fields too.
> 
> Also I wanna mention a couple things about style (not that they'd matter after my proposed changes tho): I typically use the word "kind" instead of "type" since `type` is a reserved keyword in Rust, and also I like having a space after the `//` in comments.

Updated and rebased. Really appreciate your patient help with my sloppy code. Azalea is really awesome, and I still have a lot to learn :3 Thank u

PR #263, opened 2025-10-08
Implement primitive translation arguments

### LINKED ISSUE #247: Text component primitive translation arguments
When, for example, parsing
```json
{"translate":"commands.list.players", "with": [1,2, "<players>"]}
```
with `azalea-chat` the result (`There are  of a max of  players online: <players>`) does not contain the numbers because those are deserialized to empty strings ([relevant code](https://github.com/azalea-rs/azalea/blob/58c7b3fa01bb93833a46dbb7f611e8c86f871bf7/azalea-chat/src/component.rs#L282-L286)).

Looking at the vanilla source, translation arguments are treated either as text components or primitives (in vanilla: numbers, booleans, and strings). I'd be happy to implement support for those into azalea-chat but I am not sure which approach is preferred by this project:
1. Extending the `StringOrComponent` enum by the other primitive variants (probably just `bool`, `i64` and `f64`)
2. Parsing the primitive translation arguments into `StringOrComponent::String` (or `StringOrComponent::FormattedText`)

Tert0:
Closes #247

### DIFF
--- a/azalea-chat/src/translatable_component.rs
+++ b/azalea-chat/src/translatable_component.rs
@@ -49,21 +49,81 @@ pub struct TranslatableComponent {
 }
 
 #[cfg(feature = "simdnbt")]
-fn serialize_args_as_nbt(args: &[PrimitiveOrComponent]) -> simdnbt::owned::NbtList {
-    let mut compounds = Vec::with_capacity(args.len());
+fn serialize_args_as_nbt(args: Vec<PrimitiveOrComponent>) -> simdnbt::owned::NbtList {
+    let tags: Vec<simdnbt::owned::NbtTag> = args.into_iter().map(|arg| arg.to_nbt_tag()).collect();
 
-    // TODO: do not wrap entries of homogeneous list in compounds
-    for arg in args {
-        let tag = match arg {
-            PrimitiveOrComponent::Boolean(value) => value.to_nbt_tag(),
-            PrimitiveOrComponent::Short(value) => value.to_nbt_tag(),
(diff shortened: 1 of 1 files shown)


Tert0:
> Edit: Oh, also, serialize_args_as_nbt shouldn't be wrapping the types in {"": x} when the array isn't of mixed types (so like an array of just strings shouldn't be represented as an array of compounds). This is a correctness issue but probably won't break anything significant so I won't block on it either. Probably at least worth leaving a TODO comment there though.

I'd be happy to improve this, but I am not sure how to properly implement this. I was able to check for homogeneous lists using `NbtTag::id` but I have not figured out how to convert `Vec<NbtTag>` to `NbtList` when all tags are of the same type.

mat-1:
I think it'd be best to implement that `Vec<NbtTag> -> NbtList` conversion in simdnbt rather than Azalea, but maybe it could be something like
```rs
let homogenous_list = match first_element {
    NbtTag::String(_) => list.iter().map(|e| e.into_string()).collect::<Option<Vec<Mutf8String>>>().map(NbtList::from),
    ... // repeat this for every type
}
if let Some(homogenous_list) = homogenous_list {
    return homogenous_list;
}
... // convert them all to compounds
```
(untested code, not sure if it'd work exactly like that)

mat-1:
This is fine to merge right now but if you'd like to implement that in this PR, I can wait before merging.

mat-1:
Some of this should probably be cleaned up later (by moving some things into simdnbt) but LGTM

PR #265, opened 2025-10-12
derive Message for SendGamePacketEvent

nebula161:


mat-1:
This code on its own shouldn't work (because Azalea will ignore the `Message`s), you probably want to send `SendGamePacketEvent` with `commands.trigger` instead of a `MessageWriter`.

This was changed deliberately when I updated Azalea to Bevy 0.17, since previously every `Event` could be triggered through both the `EventWriter` and `Commands::trigger` APIs, so for most trigger events I also made them support the Event API to avoid annoying errors where the events wouldn't do anything.

In Bevy 0.17 though, they finally split the two types of events (to `Message` and `EntityEvent`/`Event`) so all triggers have to use `commands.trigger` now.

nebula161:
O. Sorry, I thought it was a small oversight

mat-1:
No worries, I still appreciate the PR <3

PR #269, opened 2025-10-28
Add support for mob effects

ShayBox:


mat-1 (approved):
Thanks!

mat-1:
I made a cleanup commit in https://github.com/azalea-rs/azalea/commit/03a496fc7d952b1136cb717f2e8d1ccfb798ee64 that changes a few trivial things that I didn't want to waste your time with (since I realize that my slow review times probably aren't very enjoyable to work with lol).

Notably, I made `MobEffectData` AzBuf (and I made part of the az-protocol packet definition point to MobEffectData in az-entity) to avoid some code duplication. I also noticed that Mojang added another field to the flags that I'd missed so I took that opportunity to add that too.

Also, since `ActiveEffects` isn't wrapped in `Option` in `get_dig_speed_amplifier` and `get_effect` anymore, I moved the former into the `ActiveEffects` impl and I removed the latter (since it now does the same thing as `ActiveEffects::get_level`).

Additionally, I noticed that Azalea's jump physics could be wrong in a couple edge cases (unrelated to your changes) so I updated that too.

Thanks again for implementing this, btw :)

Issue #268, opened 2025-10-27
How use with bevy?


suprohub:
Is there any example how use this in vanilla bevy? Like this:
```rust
fn main() {
    App::new()
        .add_plugins(DefaultPlugins)
        .add_systems(Startup, setup)
        .run();
}

fn setup(mut commands: Commands) {
    todo!()
}
```

mat-1:
I wouldn't really recommend this (the main reason to do it would be to change how frequently Azalea runs the Update and GameTick schedules), but it is possible.

I haven't tested this but you should be able to do it by adding `azalea::DefaultPlugins` and `azalea::bot::DefaultBotPlugins`, then adding a system to run GameTick every 50ms before the Update (or making a custom runner like this https://github.com/azalea-rs/azalea/blob/5bc618615ec7798cf9f6779cbee8cab3bf69a898/azalea-client/src/client.rs#L627 ), and then sending `StartJoinServerEvent` to connect to the server.

suprohub:
> I wouldn't really recommend this (the main reason to do it would be to change how frequently Azalea runs the Update and GameTick schedules), but it is possible.
> 
> I haven't tested this but you should be able to do it by adding `azalea::DefaultPlugins` and `azalea::bot::DefaultBotPlugins`, then adding a system to run GameTick every 50ms before the Update (or making a custom runner like this
> 
> [azalea/azalea-client/src/client.rs](https://github.com/azalea-rs/azalea/blob/5bc618615ec7798cf9f6779cbee8cab3bf69a898/azalea-client/src/client.rs#L627)
> 
> Line 627 in [5bc6186](/azalea-rs/azalea/commit/5bc618615ec7798cf9f6779cbee8cab3bf69a898)
>  async fn run_schedule_loop( 
> ), and then sending `StartJoinServerEvent` to connect to the server.

Why you running update 60 times per sec and no as fast as can like bevy?

suprohub:
And also, are there more useful way for integrating with vanilla bevy if i cant use it directly?
Btw, you solution not works:
```
2025-10-27T17:19:47.549718Z  INFO bevy_diagnostic::system_information_diagnostics_plugin::internal: SystemInfo { os: "Linux (Arch Linux)", kernel: "6.16.4-zen1-1-zen", cpu: "12th Gen Intel(R) Core(TM) i7-12650H", core_count: "10", memory: "15.3 GiB" }
[...]

PR #271, opened 2025-10-28
Fix add player event

suprohub:
pr fixes https://github.com/azalea-rs/azalea/issues/268


suprohub:
tested right now, now all works with this fix

mat-1:
Thanks!

PR #273, opened 2025-10-30
1.21.11

mat-1:


mat-1:
untested, haven't checked for protocol changes, and haven't implemented the new data components

mat-1:
now tested in-game, should at least be mostly working

PR #281, opened 2025-11-23
Add `online-mode` Cargo feature

### LINKED ISSUE #280: Consider adding a package flag for online mode
Online mode requires https to contact Mojang servers, pulling in reqwest and rustls. Some projects don't necessarily need to authenticate as clients or verify authentications with Mojang themselves, so it might be better to move behind a flag.

mat-1:
Closes #280.

Needs testing to make sure it actually works properly and doesn't break plugins.

Would've been nice to completely disable `azalea-auth` when this feature is disabled, but currently `GameProfile` and `offline::generate_uuid` are in `azalea-auth`. Maybe they could be moved to azalea-core and azalea-crypto later?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,8 +44,10 @@ chrono = { version = "0.4.42", default-features = false }
 compact_str = "0.9.0"
 crc32c = "0.6.8"
 criterion = "0.7.0"
-crypto-bigint = "=0.7.0-rc.10"                                                # TODO: Remove when rsa is fixed.
-crypto-primes = "=0.7.0-pre.4"                                                # TODO: Remove when rsa is fixed.
+# TODO: Remove when rsa is fixed.
+crypto-bigint = "=0.7.0-rc.10"
+# TODO: Remove when rsa is fixed.
+crypto-primes = "=0.7.0-pre.4"
 derive_more = "2.0.1"
 enum-as-inner = "0.6.1"
 env_logger = "0.11.8"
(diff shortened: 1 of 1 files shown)


aparatext:
Wouldn't it also disable the AES/RSA dependencies? Offline mode uses plaintext transport I think.

mat-1:
As of a few versions ago, offline-mode servers have the option to enable encryption (not sure why tbh), so the encryption-related dependencies are unfortunately still required.

I think `rsa` dependency can be removed since it's only used for chat signing though, so thanks for pointing that out. :)

mat-1:
> Would've been nice to completely disable `azalea-auth` when this feature is disabled, but currently `GameProfile` and `offline::generate_uuid` are in `azalea-auth`. Maybe they could be moved to azalea-core and azalea-crypto later?

I moved the `offline` module to azalea-crypto in https://github.com/azalea-rs/azalea/commit/ba3a88336a074f902667d3e13471af62b431da04.

Moving GameProfile out of azalea-auth is trickier though, because moving it into azalea-core would add a relatively heavy dependency to azalea-auth (which is often used standalone).

One solution would be to significantly trim down azalea-core or add tons of feature flags, but this would just shift the problem I'm trying to solve (having lots of `#[cfg(feature="...")]` attrs). Another solution would be to make a new crate to put GameProfile in, but an entire crate just for that type doesn't feel like it would make sense. Not sure what the right solution here is.

PR #292, opened 2025-12-07
Fix EntityInteractEvent by making it an Event

### LINKED ISSUE #289: entity_interact and start_use_item do not send Interact packets
`bot.attack(entity);` will send an Interact packet (and a swing arm packet), but `bot.entity_interact(entity);` and `bot.start_use_item();` do not.

azalea clearly recognizes that the entity is *there*, because `start_use_item` will eat food if not looking at the entity, but won't eat food if it is, but there is no Interact packet sent for right clicking the entity.

EightFactorial:
Fixes #289


mat-1 (approved):
lgtm, thanks :)

PR #293, opened 2025-12-09
Add options to request Mojang sessionserver with a proxy

mat-1:
This PR introduces an `auth_proxy` field to `JoinOpts` and `ConnectOpts`, which allows clients to configure a SOCKS5 proxy for connecting to Mojang's sessionserver. It also changes the `proxy` builder functions to additionally use the given proxy for connecting to the sessionserver (which helps avoid surprising behavior on servers with `prevent-proxy-connections=true`).

The old `proxy` fields on `JoinOpts`/`ConnectOpts` were renamed to `server_proxy`, and `server_proxy` and `auth_proxy` builder functions were added as needed. I'm not sure if these field names are optimal tbh, especially auth_proxy because the proxy isn't used for *all* authentication. Maybe like "sessionserver_proxy" would be a better name, idk.

Another thing to note is that if you use a proxy, then connection pooling isn't used when `sessionserver::join` is called (because reqwest requires creating a whole new Client for every proxy). I don't think this is worth fixing.

### DIFF
--- a/azalea-client/src/client.rs
+++ b/azalea-client/src/client.rs
@@ -122,7 +122,7 @@ impl StartClientOpts {
                 address,
                 resolved_address,
                 server_proxy: None,
-                auth_proxy: None,
+                sessionserver_proxy: None,
             },
             event_sender,
         }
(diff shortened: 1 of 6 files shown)


mat-1:
ended up deciding to rename it to sessionserver_proxy :3

PR #294, opened 2025-12-11
Refactor azalea-registry

mat-1:
Clean up a lot of bad decisions in azalea-registry.

Whenever possible, to avoid breaking user code without warning, an alias type was left behind with a `#[deprecated = "moved to ..."]` attribute. These aliases will be removed in a future release.

Notable changes:
- Registry types are no longer exported in the root module of azalea-registry; "built-in" and "data" registries are now in separate modules.
- Rename `Item`, `Block`, and `DimensionType` registries to `ItemKind`, `BlockKind`, and `DimensionKind`.
- Removed "extra" registries from azalea-registry, which were hard-coded and weren't even all registries.
- Codegen all data registries (data registries are synchronized with the server as opposed to being hard-coded).
- Generate "key" enums for every data registry, so for instance there's now an `EnchantmentKey` enum that contains every vanilla variant with a fallback `Other(Identifier)` variant.
- Move `Identifier` from `azalea-core` to `azalea-registry`, mostly to avoid a cyclic dependency but also because identifiers are heavily connected to registries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ is breaking anyways, semantic versioning is not followed.
 - Refactor `RegistryHolder` to pre-deserialize some registries.
 - The handler function is now automatically single-threaded, making `#[tokio::main(flavor = "current_thread")]` unnecessary.
 - Mojang's sessionserver is now requested using the SOCKS5 proxy given in `JoinOpts::proxy`.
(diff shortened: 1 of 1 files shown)


mat-1:
i was planning on also changing entity metadata and data components to not use the unresolved data component types, but i'm not entirely sure how to implement that yet so i think i'll leave it for another day. current idea involves duplicating all the structs involved, but it'd be tricky to make a good api while also not hurting compile times too much.

PR #313, opened 2026-02-01
Fixed fuzz compilation on Windows, implemented `update_swimming`

requizm:
- Hopefully this is not going to break compilation on other operating systems 😛
- I implemented `update_swimming` via looking at the deobfuscated Minecraft code


mat-1:
The swimming implementation seems very unfinished and it's not clear to me what fuzzing change is meant to fix, since I've had other people successfully fuzz azalea-protocol on Windows before.

PR #314, opened 2026-02-01
Implemented consumeable item usage, aka foods

requizm:


mat-1:
This implementation doesn't seem correct. ServerboundUseItem should not be sent every tick, the seq should not be incremented for these packets, ServerboundPlayerAction should not be sent when an item is successfully consumed, and perhaps other things I missed. If this was AI-generated, please don't do that.

PR #315, opened 2026-02-02
Fallback to Google DNS in resolver initialization

thesalam4ik:
If the default TokioResolver builder fails, the resolver now falls back to using Google's DNS configuration. This improves reliability in environments where the system DNS configuration may be unavailable or invalid.


mat-1 (changes requested):
Thanks for the PR! This seems fine, but it should log a warning (with the `warn!` macro) with the error so the user has some feedback if they did expect it to use their system resolver.

thesalam4ik:
Done
> Thanks for the PR! This seems fine, but it should log a warning (with the `warn!` macro) with the error so the user has some feedback if they did expect it to use their system resolver.

PR #324, opened 2026-04-09
feat: make get_held_item and actions to open inventory fallible 

AfkaraLP:
The commit message basically summarises everything. We had a short conversation on the discord about possibly changing the behaviour to not panic if the client is disconnected.

There are a bunch more places where component::<T> is used, however these were the most common cases where propagating that option made sense for me personally.

However, I'd be more willing to bring this change to more places in the code (or remove it since the get_held_item and get_inventory functions now return something different from before).


AfkaraLP:
I see this has now been implemented by mat in a more thought out way, I'll close this PR. Thanks for adding this the random panics were driving me crazy

PR #326, opened 2026-04-23
fix AirSupply metadata defaulting to 0 instead of 300

wbbradley:
## Summary
`AirSupply` defaults to 0 instead of Mojang's 300. Because `ClientboundSetEntityDataPacket` only syncs non-default fields, a freshly spawned non-drowning entity reports `air_supply=0` indefinitely.

## Root cause
`codegen/lib/code/entity.py` reads defaults from burger, which returns `None` for `Int` metadata. The `default is None` branch special-cases `CompoundTag`, `*Variant`, and `VillagerData` only; `Int` falls through to `Default::default()`.

## Fix
Add `elif name == "air_supply"` in `entity.py` setting `default = "300"`.

## Test plan
- `cargo check -p azalea-entity` passes.

### DIFF
--- a/azalea-entity/src/metadata.rs
+++ b/azalea-entity/src/metadata.rs
@@ -547,7 +547,7 @@ impl Default for AbstractEntityMetadataBundle {
             currently_glowing: CurrentlyGlowing(false),
             invisible: Invisible(false),
             fall_flying: FallFlying(false),
-            air_supply: AirSupply(Default::default()),
+            air_supply: AirSupply(300),
             custom_name: CustomName(Default::default()),
             custom_name_visible: CustomNameVisible(Default::default()),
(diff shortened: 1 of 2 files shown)


mat-1:
Thanks for the PR!

I've chosen to implement this fix in Burger (https://github.com/azalea-rs/azalea-burger/commit/1742aba3e0dfc278cd4a2e489702584cd5ffac5b) instead of the codegen since it's closer to the source of the issue. I did ultimately hardcode the `300` value, though.

Here's the fix commit in Azalea: https://github.com/azalea-rs/azalea/commit/64b3c5145bb03483f593a0a8a18dd97f2c132499 (I credited you in the changelog for finding the issue)

wbbradley:
Thanks @mat-1!

Issue #87, opened 2023-04-05
`SwarmEvent` should have either a periodic or “tick” option.


Mythbusters123:
Right now, the events that a bot swarm has consists of Login, Init, Disconnect, and Chat. However, there should also be an Interval or Tick event, for when you want to do something with the swarm (like add a new account) and not have to wait for an event like Chat or Disconnect to be triggered, but rather have a period event that will go off.

mat-1:
You can add `.in_schedule(CoreSchedule::FixedUpdate)` to a system to make it run every tick.
See https://github.com/mat-1/azalea/blob/main/azalea/src/pathfinder/mod.rs#L40

Mythbusters123:
I think there should just be a `period_interval_time(Duration)` method in `SwarmBuilder`. Possibly also with a `Tick` event. I'd love to try and implement this, but I currently do not have a substantial backing with this library.

PR #328, opened 2026-05-08
Add Tick Swarm Event

Cheezer1656:
Addresses #87 by adding `SwarmEvent::Tick`, which is fired every `GameTick` from a new system in the `SwarmPlugin`


mat-1 (approved):
Thanks!

PR #330, opened 2026-05-09
Add batch packet writing

### LINKED ISSUE #284: feat: Batched write API for latency optimization
When forwarding multiple packets immediately (e.g., proxy use case), there's no way to batch writes before TCP flush. With TCP_NODELAY enabled (as in examples), each `Connection::write()` call sends a separate TCP segment.

Proposed API:

```rs
conn.write_batch(&[packet1, packet2]).await?;`
```

or

```rs
conn.begin_batch();
conn.write(packet1).await?;
conn.write(packet2).await?;
conn.flush_batch().await?;
```

This would help proxies/servers optimize connection handshakes and burst scenarios.

Cheezer1656:
Closes #284 by adding a `write_batch` function to `Connection`.

Note: the function requires owned packets, so you would do
```rs
conn.write_batch(vec![packet1, packet2]).await
```
instead of
```rs
conn.write_batch(&[packet1, packet2]).await
```
I tried making it accept references, but struggled to do so. If anyone could give a pointer on how to achieve this, I would be very grateful.

### DIFF
--- a/azalea-protocol/src/connect.rs
+++ b/azalea-protocol/src/connect.rs
@@ -223,6 +223,15 @@ where
         self.raw.write(&serialize_packet(&packet).unwrap()).await
     }
 
+    /// Write a batch of packets to the server
+    pub async fn write_batch(&mut self, packets: impl IntoIterator<Item = W>) -> io::Result<()> {
+        let serialized_packets: Vec<u8> = packets
+            .into_iter()
+            .flat_map(|packet| serialize_packet(&packet).unwrap())
+            .collect();
+        self.raw.write(&serialized_packets).await
+    }
+
     /// End the connection.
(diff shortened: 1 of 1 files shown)


mat-1 (changes requested):
Thanks for the PR! (And sorry for making you wait).

I think for this API, it'd make sense to have write_batch require that the user pass in the packet enum (rather than accepting `impl Packet<W>`), because I assume it'll be common for users to want to mix packets types in a batch.

Then, this would allow the argument (for both of the write_batch functions) to be a simple `&[W]`.

Cheezer1656:
Good idea! I didn't even realize my implementation forced all the packet types to be the same. I just pushed the updated code.

Don't worry about the wait! I'm sure you had your reasons.

mat-1 (approved):
Thanks :)

PR #333, opened 2026-06-03
fix: Frame packets correctly when writing batches

Cheezer1656:
Runs compression, length prepending, and encryption per packet when writing batches of packets instead of running it on the combined bytes of the serialized packets like before.

Sorry for introducing this issue in #330 :sweat_smile:


Cheezer1656:
The failing CI should be unrelated (same failure in 0d392ddc23fda9a5d07fa225d685ea0ff0cd171d)

mat-1 (approved):

PR #334, opened 2026-06-04
Fix OptionalNbt framing: VarInt byte length, not bool presence

etz:
﻿## Summary

Fixes the serverbound `Custom Click Action` payload framing for Minecraft 26.1.2 / 1.21.5+. The payload is now encoded as a VarInt byte length followed by unnamed NBT, with `0` meaning absent.

This PR includes protocol regression tests and fixtures under `azalea-protocol/tests/`. The live vanilla evidence is included inline below rather than committed as a new top-level docs tree.

The first commit also includes a few mechanical clippy cleanups in existing code because the exact validation command `cargo clippy -p azalea-protocol --all-targets -- -D warnings` fails on the pinned base before reaching this change otherwise.

## Affected packets

- `azalea-protocol/src/packets/config/s_custom_click_action.rs`
- `azalea-protocol/src/packets/game/s_custom_click_action.rs`

Both now use `OptionalNbt` for `payload` instead of bare `simdnbt::owned::Nbt`.

## Reproduction

Pure regression test against unpatched `86dc16c5` with a temporary packet-prefix test:

```text
thread 'optional_nbt_present_prefix_is_varint_length_not_bool' panicked at azalea-protocol\tests\optional_nbt_prefix_regression.rs:42:5:
assertion `left == right` failed
  left: [10]
 right: [105]
```

The old packet body starts the payload with raw unnamed NBT root compound byte `0x0a`; vanilla expects a VarInt byte length at that position.

Vanilla 26.1.2 localhost server repro, unpatched Azalea:

```text
[14:17:21] [Server thread/INFO]: [nbtazaleaprobe: Displayed dialog to nbtazaleaprobe]
[14:17:21] [Server thread/INFO]: nbtazaleaprobe lost connection: Internal Exception: io.netty.handler.codec.DecoderException: java.io.IOException: Packet play/serverbound/minecraft:custom_click_action (ServerboundCustomClickActionPacket) was larger than I expected, found 145 bytes extra whilst reading packet serverbound/minecraft:custom_click_action
```

Patched Azalea against the same vanilla dialog datapack:

```text
[14:14:26] [Server thread/INFO]: [nbtazaleaprobe: Displayed dialog to nbtazaleaprobe]
[14:14:26] [Server thread/INFO]: nbtazaleaprobe lost connection: Disconnected
```

There was no server-side decoder exception on the patched run.

## Byte-level evidence

Committed fixtures:

- `azalea-protocol/tests/data/optional_nbt_present_minimal.bin`: `07 0a 01 00 01 61 01 00`
- `azalea-protocol/tests/data/optional_nbt_absent.bin`: `00`
- `azalea-protocol/tests/data/optional_nbt_vanilla_dialog_probe.hex.txt`: vanilla-shaped dialog custom-action payload fixture

Live outgoing packet body, unpatched:

```text
10 61 7a 61 6c 65 61 74 65 73 74 3a 70 72 6f 62 65 0a 08 00 05 ...
```

After the `azaleatest:probe` id, the old body starts payload bytes directly with `0a`.

Live outgoing packet body, patched:

```text
10 61 7a 61 6c 65 61 74 65 73 74 3a 70 72 6f 62 65 9c 01 0a 08 00 05 ...
```

After the id, the patched body writes `9c 01`, the VarInt byte length of the 156-byte unnamed NBT payload, then the `0a` compound root.

The vanilla datapack dialog shape used for the live run was a `minecraft:notice` dialog with a text input key `field` and an action button containing `minecraft:dynamic/custom` with id `azaleatest:probe`.

## Root cause

The generated custom-click packets used `simdnbt::owned::Nbt` directly. That writes unnamed NBT bytes without the outer VarInt payload-length framing used by vanilla `FriendlyByteBuf::writeNbt` / `readNbt` for this packet in 26.1.2 / 1.21.5+.

Absent payloads can appear to work because `Nbt::None` writes `0x00`, which is also the VarInt zero-length sentinel.

## Fix

Adds `OptionalNbt`, whose `AzBuf` implementation reads/writes:

- `0x00` as absent
- non-zero VarInt byte length followed by an unnamed NBT compound payload

Then switches both known serverbound custom-click packet definitions to use it.

## Validation

Passed locally on Windows from this branch:

```text
cargo fmt --check
cargo clippy -p azalea-protocol --all-targets -- -D warnings
cargo test -p azalea-protocol optional_nbt
cargo test -p azalea-protocol
```

`cargo test -p azalea-protocol` result: 31 unit tests, 2 integration tests, and 3 doctests passed.

Live vanilla validation used `minecraft-server-26.1.2.jar`, `online-mode=false`, and an OP offline user.

## Open questions for maintainers

- Should `OptionalNbt` be renamed later to something like `LengthPrefixedNbt`? I kept the type name small and local for this fix.
- Should there be an audit for other packets where `Nbt` may have been generated for a `FriendlyByteBuf::writeNbt` / `readNbt` length-prefixed contract?


etz:
Closing and recreating from a fresh branch so the PR timeline does not retain force-pushed commits that briefly added top-level docs evidence files.

PR #337, opened 2026-06-18
fix: SlotDisplay::ItemStack uses (kind, count) wire order, not (count, kind)

wbbradley:
`ItemStackSlotDisplay` previously derived `AzBuf`, which delegates to `azalea_inventory::ItemStack`'s impl. That codec is `(count_var, item_kind_var, components)` and treats `count <= 0` as `Empty`. That is correct for the general inventory `ItemStack`, but Mojang's `SlotDisplay$ItemStackSlotDisplay` codec is `(item_kind_var, count_var, components)`, with no count-zero sentinel because the empty case is represented by the sibling `SlotDisplayData::Empty` variant.

Symptom against a vanilla server: every `RecipeBookAdd` / `UpdateRecipes` entry's result resolved to `ItemKind::from_u32(<count>)`, collapsing onto the stone-family head of the registry (Stone, Granite, Diorite, …) for any small count.

Replace the derive with a manual `AzBuf` impl matching Mojang's order. Add three tests:
  * an absolute-order decode test that hand-builds `(item_kind_var, count_var, components)` with the primitive encoders and asserts each value lands in the right field. It reads the item id from the registry enum at test time rather than from a captured payload, so it does not rot when item ids shift across Minecraft versions;
  * a round-trip that guards against codec drift between read/write (note this alone cannot catch a symmetric swap of the two leading varints, which is why the absolute-order test above exists);
  * a direct comparison vs. the inventory codec, pinning that the two leading varints are encoded in opposite order.

Affects both `ClientboundRecipeBookAdd` (via `RecipeDisplayData`) and `ClientboundUpdateRecipes` (via `SelectableRecipe::option_display`), the only consumers of `SlotDisplayData::ItemStack`.

### DIFF
--- a/azalea-protocol/src/common/recipe.rs
+++ b/azalea-protocol/src/common/recipe.rs
@@ -1,5 +1,5 @@
 use azalea_buf::AzBuf;
-use azalea_inventory::ItemStack;
+use azalea_inventory::{DataComponentPatch, ItemStackData};
 use azalea_registry::{
     HolderSet,
     builtin::{DataComponentKind, ItemKind},
@@ -88,14 +88,25 @@ pub struct OnlyWithComponentSlotDisplay {
     pub contents: SlotDisplayData,
     pub component: DataComponentKind,
 }
-
 #[derive(AzBuf, Clone, Debug, PartialEq)]
 pub struct ItemSlotDisplay {
     pub item: ItemKind,
 }
 #[derive(AzBuf, Clone, Debug, PartialEq)]
 pub struct ItemStackSlotDisplay {
-    pub stack: ItemStack,
+    pub kind: ItemKind,
+    #[var]
+    pub count: i32,
+    pub component_patch: DataComponentPatch,
+}
+impl From<ItemStackSlotDisplay> for ItemStackData {
+    fn from(item_stack_slot_display: ItemStackSlotDisplay) -> Self {
+        Self {
+            kind: item_stack_slot_display.kind,
+            count: item_stack_slot_display.count,
+            component_patch: item_stack_slot_display.component_patch,
+        }
+    }
 }
 #[derive(AzBuf, Clone, Debug, PartialEq)]
 pub struct DyedSlotDemo {


azalea-protocol/src/common/recipe.rs
```diff
@@ -93,10 +95,50 @@ pub struct OnlyWithComponentSlotDisplay {
...
+        // count_var, components)` which is distinct from the general inventory
+        // `ItemStack` codec which is `(count_var, item_kind_var, components)`
+        // and treats `count <= 0` as `Empty`. The empty case here is
+        // represented by the sibling `SlotDisplayData::Empty` variant, so this
+        // stack is always `Present` on the wire.
+        let kind = ItemKind::azalea_read(buf)?;
+        let count = i32::azalea_read_var(buf)?;
+        let component_patch = DataComponentPatch::azalea_read(buf)?;
+        Ok(Self {
+            stack: ItemStack::Present(ItemStackData {
```
mat-1: Since it's always present, I'd change `stack` to be an `ItemStackData` instead of `ItemStack` for clarity (also to be able to simplify the `azalea_write` impl).

Another approach (which would allow you to get rid of the manual AzBuf impl, and thus make the code easier to maintain) would be to just declare `ItemStackSlotDisplay` (/`ItemStackTemplate`) like this:
```rs
#[derive(AzBuf)]
struct ... {
    pub kind: ItemKind,
    #[var]
    pub count: i32,
    pub component_patch: DataComponentPatch
}
```
and having an `Into` impl for `ItemStackData`.

azalea-protocol/src/common/recipe.rs
```diff
@@ -93,10 +95,50 @@ pub struct OnlyWithComponentSlotDisplay {
 pub struct ItemSlotDisplay {
     pub item: ItemKind,
 }
-#[derive(AzBuf, Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct ItemStackSlotDisplay {
     pub stack: ItemStack,
```
mat-1: I see in Minecraft it's using an `ItemStackTemplate`, which looks like it's used in a few other places. I'm leaving it up to you as to whether you want to split that into a separate struct right now, but this should at least be mentioned in a comment so it can be found again later.

azalea-protocol/src/common/recipe.rs
```diff
@@ -121,3 +163,118 @@ pub struct WithRemainderSlotDisplay {
 pub struct CompositeSlotDisplay {
     pub contents: Vec<SlotDisplayData>,
 }
+
+#[cfg(test)]
```
mat-1: I appreciate the effort you put into writing these tests, but TBH, considering that the code that's being tested is much simpler than the tests themselves, I'd just get rid of them.

When testing protocol code, I generally find that the best approach is to capture a packet buffer from vanilla and making sure that the `azalea_read` call returns exactly what's expected (and maybe, depending on the impl, roundtripping it too). You don't need to do that for this PR, but if you still want to have a test that's how it'd typically be done in Azalea.

mat-1 (changes requested):
Thanks for finding this issue and making a PR! I've left a couple comments for things that should be changed before this is merged.

Btw, I suspect this code was AI-assisted, please don't do that. It makes the code harder to maintain and somewhat defeats the point of code reviews (which is partially to help you, as a person, be a better contributor). Apologies if this isn't the case.

azalea-protocol/src/common/recipe.rs
wbbradley: > in Minecraft it's using an `ItemStackTemplate`

Do you have a preferred link to the Minecraft sources where I could investigate that? Thank you.

wbbradley: I could not find a reference to `ItemStackTemplate` anywhere, so for now I have just left it as a fix to `ItemStackSlotDisplay`. I'd gladly add a comment but I don't know how to write such a comment without understanding what `ItemStackTemplate` is or where it's used, etc...

azalea-protocol/src/common/recipe.rs
wbbradley: Thanks, I went with this approach.

azalea-protocol/src/common/recipe.rs
wbbradley: Removed.

PR #342, opened 2026-06-23
feat: Elytra

commandblock2:
Edit: This implements elytra physics. For elytra, it's basically that 
- the client to enter the gliding state
  - on entering, the client toggles the metadata `FallFlying` flag
  - then client sends a command packet to server immediately saying it has began gliding
  - then the client uses the `FallFlying` flag for travel in that same tick
- on touch down the server tells the client to stop
  - server sends a update to the `FallFlying` flag
  - client updates it.
- another case: client flies onto a claimable, client stops locally in `travel()`

There are some cases that's not yet supported in this PR, eg. Slow Falling will block elytra activation, fireworks consumption.

What's tested right now against Grim are
- continuous flying over > 2k blocks
  - flags on ~colliding with world border (but I think that's a todo item in the repo already?)~ what I really meant was flying into not yet loaded chunks, though it probably should also flag on world border
- continuous landing
  - works on ice and normal blocks
  - ~flags in an edge case (more on that later)~
- sprint jumping under a 2 blocks high roof without an elytra (discovered a Grim flag irrelevant to this change, see later comments) (to test and found no obvious regression)
- flying into into a orthogonal/parallel(to the flying direction) wall of ladder and exit gliding(can't care to record a new video)
- tested with the following bot
  
  ```rust
  use azalea::{WalkDirection, prelude::*};
  
  const SERVER: &str = "localhost";
  const BOT_NAME: &str = "walker1";
  const JUMP_EVERY_TICKS: u64 = 2;
  
  #[tokio::main]
  async fn main() -> AppExit {
      let account = Account::offline(BOT_NAME);
  
      ClientBuilder::new()
          .set_handler(handle)
          .start(account, SERVER)
          .await
  }
  
  #[derive(Clone, Component, Default)]
  struct State;
  
  async fn handle(bot: Client, event: Event, _state: State) -> eyre::Result<()> {
      match event {
          Event::Spawn => {
              bot.set_direction(180.0, 0.0)?;
              bot.walk(WalkDirection::Forward);
          }
          Event::Tick => {
              bot.sprint(azalea::SprintDirection::Forward);
              let ticks = bot.ticks_connected();
              if ticks > 0 && ticks % JUMP_EVERY_TICKS == 0 {
                  bot.jump();
              }
          }
          _ => {}
      }
  
      Ok(())
  }
  ```
  
[elytra.webm](https://github.com/user-attachments/assets/c0bb92b9-4d3e-4aa8-83b0-0cb412bf4c61)

The flagging(now solved) is at the end of this demo video. This basically only happens when we have 2 blocks high space and _my speculation is the bot is trying to re-enter gliding state, using the bounding box of gliding or something but Grim thinks the the player has another Pose like Crouching_. If you are interested in this, there is also a Grim SuperDebug log, coming from a modified Grim version with more output info. https://paste.grim.ac/KGxo8jFTbk. (This case can be kinda tricky if you want to reproduce it, doesn't always happen in a 2 block high ceiling).

Edit: also our math showed more float error than vanilla's (vanilla has a E-13 to E-16 error but ours is almost E-6 to E-7 when gliding, Grim doesn't flag us for that though), not sure if that's expected.

Edit: btw I have fixed travel and only require FallFlying to be optional so that the tests won't fail because of physics not working.


commandblock2:
https://github.com/user-attachments/assets/b953b158-ad39-4c99-bceb-4f7b237f4c19

Update: this now can properly change it's pose and won't flag in this case. But it seemed to introduce a bit more problem elsewhere (flags and stuck intermittently when sprint jumping under a 2 block roof without elytra on bot).

Edit: I was testing between commits before I change things vs this branch so I do know that's what this change introduced. This put dimension and bounding box update to GameTick and is ran after travel(similar to what vanilla does) but before we send input or movement packets to the server, so it's pretty much expected to change things, but at this point I do not quite understand fully why this consequence is here and what else would need to be changed

Edit: flags for this https://paste.grim.ac/EmrFIeloDv

commandblock2:
It turns out the 0.2 block flag was not introduced by this feature, azalea had this flag before I started working on elytra related stuff, it is just that I have broke tp packet handling and Grim's every reset on Grim causes yet another reset so it was much more significant. Will update with videos later.
Edit: I said this because I have a definitive reproduction(not a minimal one though)

Edit: this maybe a Grim false flag though, we will have to see that later.

// this one is current commit

https://github.com/user-attachments/assets/fbb3f2e4-c04e-4947-b5bb-1c49000b7c28

// this one is when this branch is created
https://github.com/user-attachments/assets/c0ab9240-5038-403a-b160-538ec6c89a21

commandblock2:
Update: This should now be ready for review. and it should not introduce more flags. The reason that it flagged was because I removed the upcomming updates for dimensions and boundingbox on tp packet and it generally flagged more on both Grim sent resets and /tp command. Not sure what's the exact mojmap name for the tp packet, should probably be the same thing of S08 or something back in the 1.8 searge mapping.

Edit: this is only tested on 26.1 but not 26.2, and some extra info, [I think 26.3 may have changed the elytra physics a bit](https://bugs.mojang.com/browse/MC/issues/MC-305535).

azalea-physics/src/travel.rs
```diff
@@ -225,6 +229,60 @@ fn travel_in_fluid(ctx: &mut MoveCtx) {
...
+            let convert = movement.y * -0.1 * lift_force;
+            movement += Vec3::from((
+                look_angle.x * convert / look_horizontal_length,
+                convert,
+                look_angle.z * convert / look_horizontal_length,
+            ));
+        }
+
+        if lean_angle < 0.0 && look_horizontal_length > 0.0 {
+            let convert = move_horizontal_length * -f64::from(lean_angle).sin() * 0.04;
```
mat-1: This should use `azalea_core::math::sin` (which is less precise than f64::sin but matches the behavior of `Mth.sin`). This doesn't apply to the `.cos()` above because Mojang used `Math.cos` for that one, lol.

azalea-physics/src/travel.rs
```diff
@@ -225,6 +229,60 @@ fn travel_in_fluid(ctx: &mut MoveCtx) {
...
+        if look_horizontal_length > 0.0 {
+            movement += Vec3::from((
+                (look_angle.x / look_horizontal_length * move_horizontal_length - movement.x) * 0.1,
+                0.0,
+                (look_angle.z / look_horizontal_length * move_horizontal_length - movement.z) * 0.1,
+            ));
+        }
+
+        ctx.physics.velocity =
+            Vec3::from((movement.x * 0.99, movement.y * 0.98, movement.z * 0.99));
```
mat-1: These are 32-bit floats in vanilla, so they should be written like `0.99f32 as f64`.

(also, noting that it'd be a bit cleaner if you used Vec3's `.multiply`).

mat-1 (changes requested):
Code looks maintainable and I can confirm that it works. Well done!

I spotted a couple correctness issues (posted as separate review comments), those should be fixed before this is merged.

Also, here's some minor nitpicks/notes (you don't have to fix any of these):
- I'd rename `fallflying` to `fall_flying` for consistency.
- "vanilla convert to double first, not sure if it has to be this" it does have to be like this, otherwise we lose precision. :(
- You can do `Vec::new(x, y, z)` instead of `Vec::from((x, y, z))`.
- `movement.y += y;` would be easier to read than `movement += Vec3::new(0, y, 0);`.
- The comment on `EntityGeometryUpdateSystems` could be a doc comment.
- "vanilla first set to true then set to false again, quite confusing" I believe vanilla's doing this to mark it as dirty to force servers to resend the entity data packet? yeah it's definitely weird though.
- I'd change the thing in the README to something more commonly used, perhaps sprint-swimming.

azalea-physics/src/travel.rs
commandblock2: Fixed, never knew that mojang even have their own fast math

azalea-physics/src/travel.rs
commandblock2: that's also fixed now

commandblock2:
Thanks for your time and effort for this review! I think the 2 correctness issue
- should use `azalea_core::math::sin` when Mojang uses their `Mth.sin`
- f32 literal widened to f64 on purpose.

are fixed now and elytra movement goes nearly go E-15 to E-16 in Grim's debug.

And the rest of the notes should be fixed as well, they probably are also kinda important for code quality.

mat-1 (approved):
Thanks!

PR #344, opened 2026-06-27
feat: Added check for crouching when moving in Water

ErriicGit:
The bot now moves down as expected when crouching in water.
Its pretty much exactly like the minecraft implementation found in the LocalPlayer class.
Grim was also happy with it.

### DIFF
--- a/azalea-physics/src/lib.rs
+++ b/azalea-physics/src/lib.rs
@@ -9,25 +9,28 @@ pub mod travel;
 
 use std::collections::HashSet;
 
-use azalea_block::{BlockState, BlockTrait, fluid_state::FluidState, properties};
+use azalea_block::{fluid_state::FluidState, properties, BlockState, BlockTrait};
 use azalea_core::{
     math,
     position::{BlockPos, Vec3},
     tick::GameTick,
 };
 use azalea_entity::{
-    ActiveEffects, Attributes, EntityKindComponent, HasClientLoaded, Jumping, LocalEntity,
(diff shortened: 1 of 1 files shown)


azalea-physics/src/lib.rs
```diff
@@ -123,6 +125,10 @@ pub fn ai_step(
             physics.z_acceleration *= 0.98;
         }
 
+        if crouching == Some(&Crouching(true)) && physics.is_in_water(){
```
mat-1: Vanilla uses `this.input.keyPresses.shift()`, which is technically not the same as `Crouching` (because there are cases where we can be crouching without pressing shift). This should be checking `.trying_to_crouch` on the `ClientMovementState` component instead.

mat-1 (changes requested):
Thank you very much! There's one correctness issue I spotted, but this otherwise looks good :).

BTW, I'd recommend running `cargo fmt` before committing.

ErriicGit:
Thank you for the review.
I must have mixed it up, because you call bot.set_crouching(true), but it doesn't actually directly set the value.
But it should be fixed and formatted now and is still running fine.

mat-1 (approved):
I noticed another small correctness issue (the 0.04 should be a float), but this was kinda my fault since I made the same mistake in `jump_in_liquid`, so I just made a quick commit to fix it.

I've now merged your update here: https://github.com/azalea-rs/azalea/commit/a8959f6264b2d6bc15d64db22cd957797818cef2 (I accidentally merged it into the wrong branch first, oops)

PR #345, opened 2026-07-03
Add azalea-reflection-proxy to README

RaymondShell:


mat-1:
I'm holding off on merging this (for now, at least) because your project seems largely vibe-coded. My reasoning for wanting to avoid this is that AI generated code is against the spirit of what I want to promote (you will significant limit your learning if you don't force yourself to think through everything), and it tends to be unmaintainable and unidiomatic (and thus is not a good reference for people wanting to learn).

PR #346, opened 2026-07-03
Improve performence of block trait

urisinger:
## Description

Optimizes the conversion from `BlockState` to `BlockTrait` by replacing the monster match statment with a static table of `&'static dyn BlockTrait`.

* Adds `BlockState::to_trait()` for easy reference conversion.
* Adds a fallback method on `BlockTrait` to get a `Box<dyn BlockTrait>` when ownership is required.
* Adds benchmarks to track conversion performance.

## Benchmarks

*Note: These benchmarks use a randomized selection of blocks. Real-world performance may vary depending on cache locality and access patterns.*

```text
&dyn BlockTrait (Reference)
  Before: 390.14 µs  ->  After: 1.2546 µs  (-99.68% / ~310x faster)

Box<dyn BlockTrait> (Heap)
  Before: 396.05 µs  ->  After: 41.820 µs  (-89.44% / ~9.5x faster)

```

## Binary Size Impact

The file size of the benchmark binary increased by **7 MB** due to the addition of the static table. Note that this impact has only been measured on the standalone benchmark program and has not yet been tested on a full Azalea build.

## AI Usage
AI was used to write the benchmark code and the pr description, all other code was written by hand.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -296,6 +296,8 @@ dependencies = [
  "azalea-block-macros",
  "azalea-buf",
  "azalea-registry",
+ "criterion",
+ "rand 0.10.2",
 ]
 
 [[package]]
--- a/azalea-block/Cargo.toml
+++ b/azalea-block/Cargo.toml
@@ -11,5 +11,13 @@ azalea-block-macros.workspace = true
 azalea-buf.workspace = true
 azalea-registry.workspace = true
 
+[dev-dependencies]
+criterion.workspace = true
+rand.workspace = true
+
+[[bench]]
+name = "block"
+harness = false
+
 [lints]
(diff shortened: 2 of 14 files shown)


EightFactorial (dismissed):

mat-1 (approved):
LGTM, thanks! I'm still wondering how this may affect compile times, but it doesn't seem too significant and there's other parts of Azalea that are much worse anyways. Merging. :)

urisinger:
> LGTM, thanks! I'm still wondering how this may affect compile times, but it doesn't seem too significant and there's other parts of Azalea that are much worse anyways. Merging. :)

Should not be too bad, i think when testing it added 5 seconds

PR #353, opened 2026-07-17
docs: Fix readme link size for accessibility

PseudoResonance:
I was just trying to read this on mobile and had a hard time clicking the tiny asterisks. I'd appreciate it if you keep all links longer than a couple characters!

Change it to whatever you want though, no need to merge this in particular if you'd rather something else. Thanks for the great project!


mat-1 (approved):
Thanks :)

PR #371, opened 2026-07-26
feat: Main supporting blockpos

commandblock2:
For issue https://github.com/azalea-rs/azalea/issues/350
~~Not tested, DO NOT MERGE (yet)~~

This PR introduces `main_supporting_blockpos` tracking to match vanilla, including
- updating `main_supporting_blockpos` when doing `set_on_ground` (changes the signature of `set_on_ground`, because vanilla updates those two)
  - call site within `move_colliding`, it additionally updates the bounding box after position is updated(need review for maintenance, and that update was missing all the time before this).
  - call site for packet handling, it now uses the updated bounding box. However this leaves the later `update_bounding_box` system redundant.
- uses `main_supporting_blockpos` for the `on_pos` and etc.. so that they can use the correct blockpos for determining support

Tested with a example bot, knocked down a platform, with different starting position to test edge case of landing tick moves the bot out side of the ice platform (no contact), see later posts with videos to know what tests are done.

PR #374, opened 2026-08-01
feat: updating main supporting blockpos with separate systems

commandblock2:
I think this code is cleaner to read if we don't want it to look exactly like vanilla. I think I would recommend this branch over the old pr after looking at this code finally, even if this looks less resemblance to vanilla logic and maybe thus harder to debug if you debug it by looking at the diff between sources.

This is too tested with all the tests applied in the [previous branch](https://github.com/azalea-rs/azalea/pull/371), with `GrimAC (2.3.74-d54b49b)` on paper 26.2.

This PR includes 
- 2 new systems, `update_main_supporting_block_pos` and `update_falling_distance` (was in `move_colliding`), and inserted a additional pass of `update_bounding_box`, after `travel`.

  This is because in vanilla 
  - `Entity.setPos` means `Entity.setBoundingBox` also gets called 
  - `Entity.setOnGround` and `Entity.setOnGroundWithMovement` means `Entity.checkSupportingBlock` also gets called.
  
  Those two conditions were not modeled as close as now is in azalea. (`travel` should update bounding box, and `update_main_supporting_block_pos` should be ran after `travel` and use that updated bounding box).
  
- a change to existing `BlockCollisionsState` api, because vanilla counterpart allows to select either a `BlockPos` or a `VoxelShape`, now we also do the same thing. Changes `&mut Vec<>` pointer to a function pointer and calls, may be slower? Haven't done profiling but think should have minimal impact.

- Other related changes so that `get_on_pos` can make use of the newly added `main_supporting_blockpos` when doing movement physics.

### DIFF
--- a/azalea-client/src/plugins/attack.rs
+++ b/azalea-client/src/plugins/attack.rs
@@ -30,7 +30,7 @@ impl Plugin for AttackPlugin {
                     update_attack_strength_scale.after(PhysicsSystems),
                     // in vanilla, handle_attack_queued is part of `handleKeybinds`
                     handle_attack_queued
-                        .before(super::movement::update_pose)
+                        .before(super::movement::send_sprinting_if_needed)
                         .before(super::tick_end::game_tick_packet),
                 )
                     .chain(),
--- a/azalea-client/src/plugins/movement.rs
+++ b/azalea-client/src/plugins/movement.rs
@@ -69,11 +69,9 @@ impl Plugin for MovementPlugin {
                         .before(ai_step)
                         .before(azalea_physics::fluids::update_in_water_state_and_do_fluid_pushing),
                     send_player_input_packet,
-                    update_pose.before(EntityGeometryUpdateSystems),
+                    update_pose.after(travel).before(EntityGeometryUpdateSystems),
                     send_sprinting_if_needed
-                        .after(azalea_entity::update_in_loaded_chunk)
-                        .after(travel)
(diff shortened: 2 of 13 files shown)


commandblock2:
```
---- azalea-core/src/codec_utils.rs - codec_utils::flatten_array (line 28) stdout ----
error: expected item after attributes
  --> azalea-core/src/codec_utils.rs:29:1
   |
29 | #[serde(serialize_with = "flatten_array")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an item after this
```

This did not happen with my older nightly tool chain, but after updating my tool chain(with `rustup update nightly`) to 1.99 it reproduces this, seems that `no_run` will try to compile it as well.

commandblock2:
It turns out that if we were to further properly implement `blockSpeedFactor` and `restituteMovementAfterCollisions`, we'd either need to choose the following

1. breaking up `travel` into different systems, to system(s) thats before `move_colliding`, `move colliding` as a system, `update_supporting_block`, and system(s) thats after `move_colliding`
2. revert this and go to the way that the previous PR does it.

Because, those two system both modifies the velocity (which is used by later part of `travel`, and both of them relies on this supporting blockpos to be updated in place instead of a system ran later).

And either of them requires a very big change for both. So this probably should not be merged, at least not in the state right now. But I will leave this here for others to reference and I will continue to try the first option.

PR #378, opened 2026-08-25
Fix Rust 1.100.0 warnings, force older trait solver

EightFactorial:
Hopefully `generic_const_exprs` is added back soon...

Until then the solver must be set with `-Znext-solver=coherence`.
Doing it in-repo only effects azalea dev, other projects must set it themselves.

Edit: Seems like `min_generic_const_args`/`generic_const_args` is a "successor"?
Or at least a replacement. Either way, the Rust doc examples for it don't seem to work yet.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,8 @@
 [build]
-rustflags = ["--cfg", "docsrs_dep", "-Znext-solver=coherence"]
+rustflags = [
+  "--cfg", "docsrs_dep",
+  "-Znext-solver=coherence" # This is temporary until the new trait solver supports `min_const_generic`s
+]
 rustdocflags = [
     "--html-after-content",
     "docs-rs/trait-tags.html",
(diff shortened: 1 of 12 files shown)


.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
 [build]
-rustflags = ["--cfg", "docsrs_dep"]
+rustflags = ["--cfg", "docsrs_dep", "-Znext-solver=coherence"]
```
mat-1: I'd put a comment explaining why we need this for now, since we'll want to remove it eventually.

azalea/src/pathfinder/goals.rs
```diff
@@ -55,16 +55,7 @@ fn xz_heuristic(dx: f32, dz: f32) -> f32 {
...
-    let straight;
-
-    if x < z {
-        straight = z - x;
-        diagonal = x;
-    } else {
-        straight = x - z;
-        diagonal = z;
-    }
+    let (diagonal, straight) = if x < z { (z - x, x) } else { (x - z, z) };
```
mat-1: This flips the variable assignments. Kinda my fault since the variables were defined in the opposite order that they're used, oops.

mat-1 (changes requested):

azalea/src/pathfinder/goals.rs
EightFactorial: Lol,  I could make that a lot more readable too

mat-1 (approved):

PR #376, opened 2026-08-11
refact&feat: support for slime/bed(bouncing)/honey/soul sand/ice(edge)/web/bush

commandblock2:
Solves these two issues
https://github.com/azalea-rs/azalea/issues/354
https://github.com/azalea-rs/azalea/issues/350

Relies on this https://github.com/azalea-rs/azalea/pull/374, but I think I am gonna keep it open for a little while

Add support for following block

- slime & bed(bouncing), 
  - bouncing, both uses bounciness property of block behavior

    https://github.com/user-attachments/assets/4b6c9f25-2f00-446b-9b19-43a91a6462b0

  - slowdown for slime, uses `Entity.stepOn` in vanilla, and it's the only one that uses stepOn with a effect on client side
  
    https://github.com/user-attachments/assets/95893806-6c8f-4897-9dba-0c9aaf683d89

- honey block
  - sliding (`check_inside_blocks`)
 
    https://github.com/user-attachments/assets/4069fe29-64ea-44be-80ea-7c4eb88dbf33

    Note that the flag may be re-produced with a vanilla client, relevant Grim issue at https://github.com/GrimAnticheat/Grim/issues/2829. Bubble column also tested since they share same path
  - slowdown (`apply_speed_factor`)
  
    https://github.com/user-attachments/assets/e7cee311-f41a-4381-ac92-83125bff4c5b

- soul sand
  - slowdown (`apply_speed_factor`)
  
    https://github.com/user-attachments/assets/fb6801c6-1f98-4569-96f4-e8822475fdc5

- ice (edge caases)
  - sliding off a platform by knockback
    
    https://github.com/user-attachments/assets/867332f6-f835-43e3-894f-ace8e9df68c1

  - landing on the edge of the ice block

    https://github.com/user-attachments/assets/dfba7378-9683-45fa-9094-5fbe5d5da428

- web/bush
  - slowdown (stuck multiplier)
  
    https://github.com/user-attachments/assets/cbbfd61d-3cca-46a4-801f-b851465a0479

    https://github.com/user-attachments/assets/083cdcab-3142-4d90-bf1d-2469cda9b15a

    https://github.com/user-attachments/assets/9f51c250-8037-4472-8612-7e8c5ee104ce

Kinda idk what to say to discover that vanilla actually have 3 different ways at different time to apply slowdown to the player.

Changes that made this possible (wouldn't call this a good refactor because the code did seem to be much cleaner, hope any who reviews give useful tips):

- `travel` is broken into different systems after `move_colliding`
[...]

PR #379, opened 2026-08-26
Fix: corner case for fluid height (should be 1 when there is the same kind of fluid above)

commandblock2:
Before:

https://github.com/user-attachments/assets/734196ae-4d84-47f3-8e69-995e92a2e81d

After:

https://github.com/user-attachments/assets/1510ef8a-78ee-473c-a8e1-196c8131ea82

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,8 @@
 [build]
-rustflags = ["--cfg", "docsrs_dep", "-Znext-solver=coherence"]
+rustflags = [
+  "--cfg", "docsrs_dep",
+  "-Znext-solver=coherence" # This is temporary until the new trait solver supports `min_const_generic`s
+]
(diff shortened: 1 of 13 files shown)


commandblock2:
based on [EightFactorial](https://github.com/EightFactorial)'s [branch](https://github.com/azalea-rs/azalea/pull/378) so that it may be easier to test for anyone who wants to test.

---

Additionally for a shameless self advertising, I personally have a [branch](https://github.com/commandblock2/azalea/commits/swimming-and-water-fix) with swimming.

The branch does have 
- swimming physics (without minor horizontal collision modeling) 
- and a additional fix when entering water


that is built on top of my [other branch not merged yet](https://github.com/azalea-rs/azalea/pull/376), so I would not put the PR here just yet, if you are interested in swimming you could try this branch.

PR #380, opened 2026-08-29
Migrate from `generic_const_exprs` to `generic_const_args`

EightFactorial:
Actually fixes Rust 1.100 :)


EightFactorial:
Hm, this function doesn't seem right
https://github.com/azalea-rs/azalea/blob/153c90aa5570b3a94b909e82c1c0b60a90fff5db/azalea-protocol/src/packets/game/c_commands.rs#L246

mat-1 (dismissed):

mat-1:
> Hm, this function doesn't seem right
> https://github.com/azalea-rs/azalea/blob/153c90aa5570b3a94b909e82c1c0b60a90fff5db/azalea-protocol/src/packets/game/c_commands.rs#L246
> 

Oh, you're right. Would probably be best to double-check what vanilla does, but that 4 should indeed almost certainly be a 5.

mat-1 (approved):

PR #381, opened 2026-09-03
Remove AttackQueued unconditionally, not just on success

Cmothersell:
handle_attack_queued only removed the AttackQueued component on the success path - if the EntityIdIndex lookup failed (the target died before this tick processed the queued attack, which is inherently the common case for a killing blow specifically, since that attack is queued while the target is still alive), the component was left attached to the client entity. Since this system runs every GameTick and its query matches on AttackQueued's mere presence, that stale component kept re-matching - and re-warning - on every single subsequent tick, indefinitely, until some unrelated future attack() call happened to overwrite it.

Observed live: a wall of "tried to attack entity ... which isn't in our EntityIdIndex" warnings continuing indefinitely after every kill.


mat-1 (dismissed):
Good catch, thanks! Will merge in a bit.

I have one nitpick (which I won't block on), which is that the comment is quite verbose. For the PR description it's fine, but for the code itself a single-sentence comment explaining why it's like that would fit better (or no comment at all, since it's already kinda self-explanatory).

Cmothersell:
Good call. I like to document as I write and build PR notes from that.

mat-1 (approved):

